PR #1, opened 2017-01-26
Updates to config / yaml structure and push / deploy commands

izakp:
Updates:

- Adds `hokusai/common.yml` and `hokusai/staging.yml`
- `hokusai build` command uses `hokusai/common.yml` - this allows for passing through environment variables as build args (see current use in Currents)
- `hokusai stack_up / stack_down` now take a context (staging or production) and implicitly look for the `{context}.yml` to create the stack, separating staging and production stacks
- `hokusai dev / test` commands now also implicitly look for `development.yml` / `test.yml` respectively
- `hokusai push` now accepts only one argument - the tag to push to the repo.  It should be called with the git SHA of the build.  The `--test-build` option should be called from CI and pushes the latest image output by the `hokusai test` command.  Otherwise it pushes the latest image output by the `hokusai build` command
- `hokusai deploy` now is invoked with `{context}` and `{tag}` as positional arguments.  Before the deploy, it updates the `{context}` tag (i.e. `staging` / `production`) to point to the tag it then deploys, keeping it up-to-date with the latest deployment
- merges `hokusai stack_up` / `hokusai stack_down` commands into `hokusai stack` command with subcommands `up / down / status`

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,12 +72,16 @@ Required options:
 
 #### Launching a stack
 
-`hokusai stack_up {CONTEXT}` launches the stack defined in `hokusai/production.yml' for the given Kubernetes context
+`hokusai stack up {CONTEXT}` launches the stack defined in `hokusai/{CONTEXT}.yml' for the given Kubernetes context
 
 #### Deleting a stack
 
-`hokusai stack_down {CONTEXT}` deletes the stack defined in `hokusai/production.yml' for the given Kubernetes context
+`hokusai stack down {CONTEXT}` deletes the stack defined in `hokusai/{CONTEXT}.yml' for the given Kubernetes context
+
+#### Checking a stack status
+
+`hokusai stack status {CONTEXT}` prints the stack status defined in `hokusai/{CONTEXT}.yml' for the given Kubernetes context
 
 #### Rolling deployments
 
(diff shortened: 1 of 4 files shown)


izakp:
Note:  tested with latest docker for mac, which includes:

docker version 1.13.0
docker-compose version 1.10.0, build 4bd6f1a
docker-py version: 2.0.1

CircleCI does not support these versions yet and should be pinned to docker version 1.10.0 / docker-compose version 1.9.0

izakp:
Also FYI:  If, while running `hokusai dev` / `hokusai test` you get an error like:

```
ERROR: for delta  No such image: sha256:575645359b466cc72febe72d710d5ced67877a986fcc7a99a5d8c7217b28bf9a
ERROR: Encountered errors while bringing up the project.
```

You probably cleaned up your local docker image registry and deleted those containers.  Since hokusai is not removing containers any more (by calling `docker-compose stop` rather than `docker-compose kill`) it looks for already created containers to bring back up.  You can resolve this issue by running `docker-compose -f hokusai/development.yml rm && docker-compose -f hokusai/test.yml -p ci rm` to clean stale references to container images

Might add this as a `hokusai clean` command in the future

bhoggard:
I don't understand how flags like `with_memcached` get set, so we should explain that in the README. I will merge now so I can test with currents.

PR #3, opened 2017-02-14
improve concision and scan-ability of readme. add about section.

cavvia:
There's a number of formatting fixes here too. 

In general i think there's no need to discuss cli command options in depth here as using the `--help` option will be far more practical and it complicates the readme. Just an overview of the avail commands is good enough.


izakp:
Thanks! 👍

PR #4, opened 2017-02-15
Big refactor - split hokusai/__init__.py into separate files

izakp:
`hokusai/__init__.py` was getting too large - split into individual files in `hokusai/commands/` and refactored imports to be explicit


izakp:
Also drop `add_secret` command and replace with `hokusai secrets pull` / `hokusai secrets push` - pulls secrets from the server and writes yml out to `hokusai` directory, allowing for local editing.  `push` then uploads it to the server.

Add `config` command that functions in this way as well.

hokusai/commands/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+from hokusai.commands.deploy import deploy
+from hokusai.commands.development import development
+from hokusai.commands.images import images
+from hokusai.commands.init import init
+from hokusai.commands.pull import pull
+from hokusai.commands.push import push
+from hokusai.commands.run import run
+from hokusai.commands.secrets import pull_secrets, push_secrets
+from hokusai.commands.stack import stack_up, stack_down, stack_status
+from hokusai.commands.test import test
```
cavvia: much nicer logical organisation

hokusai/commands/check.py
```diff
@@ -0,0 +1,140 @@
...
+import urllib
+import shutil
+import getpass
+from distutils.dir_util import mkpath
+
+from subprocess import check_output, CalledProcessError, STDOUT
+
+from hokusai.common import print_red, print_green
+
+def check(interactive):
```
cavvia: huge function - as this is an interactive cli process, i'm not sure how easy it is to break out helpers or more sub functions but worth thinking about. e.g. you have AWS, kubectl and docker installation steps below, could be broken out potentially.

hokusai/commands/check.py
```diff
@@ -0,0 +1,140 @@
...
+                             % (kubectl_version, platform), os.path.join('/tmp', 'kubectl'))
+          os.chmod(os.path.join('/tmp', 'kubectl'), 0755)
+          shutil.move(os.path.join('/tmp', 'kubectl'), os.path.join(install_to, 'kubectl'))
+          mkpath(os.path.join(os.environ.get('HOME'), '.kube'))
+          print_green("kubectl installed")
+          print_green("Now install your organization's kubectl config to ~/.kube/config")
+        except Exception, e:
+          print_red("Installing kubectl failed with error %s" % e.message)
+
+  if os.path.isfile(os.path.join(os.environ.get('HOME'), '.kube', 'config')):
```
cavvia: lots of repetition in the following if else clauses which could be designed out, e.g. by iterating over an array of file paths and checking each in turn.

hokusai/commands/config.py
```diff
@@ -0,0 +1,74 @@
...
+    print_red("Failed to pull configmap hokusai/%s-config.yml" % context)
+    return -1
+
+  print_green("Pulled configmap hokusai/%s-config.yml" % context)
+  return 0
+
+def push_config(context):
+  HokusaiConfig().check()
+
+  if not os.path.isfile(os.path.join(os.getcwd(), 'hokusai', "%s-config.yml" % context)):
```
cavvia: you could do with a helper for this general pattern, e.g. a file checking function that takes a relative path. used in the last command too.

hokusai/commands/config.py
```diff
@@ -0,0 +1,74 @@
+import os
+
+from collections import OrderedDict
+from subprocess import check_output, CalledProcessError, STDOUT
+import yaml
+
+from hokusai.config import HokusaiConfig
+from hokusai.common import print_red, print_green
+
+def pull_config(context):
```
cavvia: Inline comments describing the command and args would be useful to devs looking at the code. Though obviously the comments are all compiled in `bin/hokusai` to be accessed by users of the tool, inline would be good for contributing devs.

hokusai/commands/init.py
```diff
@@ -0,0 +1,169 @@
...
+
+from jinja2 import Environment, PackageLoader
+env = Environment(loader=PackageLoader('hokusai', 'templates'))
+
+from hokusai.config import HokusaiConfig
+from hokusai.common import print_green, build_service, build_deployment, YAML_HEADER
+
+def init(project_name, aws_account_id, aws_ecr_region, framework, base_image,
+          run_command, development_command, test_command, port, target_port,
+            with_memcached, with_redis, with_mongo, with_postgres):
```
cavvia: monster of a function signature! this command could probably benefit from some state, though i'm aware 'click' encourages this more functional approach.

hokusai/commands/config.py
izakp: 👍

hokusai/commands/init.py
izakp: Right - the init command still bothers me.  It could me made interactive as it's only meant to be run when adding hokusai to a project?

hokusai/commands/check.py
izakp: Agreed

cavvia:
Extracting all the commands out makes this a lot more approachable from the pov of a contributing dev or anyone wanting to inspect the internals of a given command. :+1:

hokusai/commands/config.py
izakp: Yea [docstrings](https://www.python.org/dev/peps/pep-0257/) would be a good addition

hokusai/commands/secrets.py
```diff
@@ -0,0 +1,76 @@
...
+
+import yaml
+
+from hokusai.config import HokusaiConfig
+from hokusai.common import print_red, print_green
+
+def pull_secrets(context):
+  config = HokusaiConfig().check()
+
+  try:
```
izakp: Another common pattern which should be extracted into a shared function

izakp:
@bhoggard added options to `console` and `run` commands that will in application config and secrets and inject those key/value pairs as environment variables (defaults to true)

PR #5, opened 2017-02-22
add elixir to init command

izakp:


izakp:
Let's go through this together!

PR #7, opened 2017-02-24
Updates to verbosity, command outputs

izakp:
- Most hokusai commands now accept a `-v` / `--verbose` flag
- Adds `hokusai promote` to copy the deployment from one context to another. Use: `hokusai promote production staging` promotes production from staging
- Updates to `deploy` command to handle multiple containers in a deployment (currently affects Currents)

Coming soon! Support for multiple deployments in a given app so as to scale web / async workers independently by using Label Selectors https://kubernetes.io/docs/user-guide/labels/

cc @bhoggard @cavvia

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ Required options:
 ### Deployment
 
 * `hokusai deploy` - Update the Kubernetes deployment to a given image tag.
+* `hokusai promote` - Update the Kubernetes deployment in a given context to match the deployment in another context
 
 ### Running a console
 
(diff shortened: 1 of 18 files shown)


bhoggard:
I feel that `hokusai promote staging production` seems more idiomatic, as in staging => production.

izakp:
@bhoggard agreed, now that I take another look at it.  Will change that.

joeyAghion:
That was my initial reaction about the staging/production order too.

PR #13, opened 2017-03-10
Fix deployments and shell

izakp:
- Uses a `kubectl patch` for deployments: `hokusai deploy` / `hokusai promote` adding the "deploymentTimestamp" label to the deployment metadata, which always triggers a rollout, even if the pod image is the same

- Adds "ImagePullPolicy": "Always" to `hokusai console` and `hokusai run` commands, ensuring newest image is always run for the given tag

Update your hokusai installation with: `git pull upstream master` / `pip install --upgrade .`

cc @joeyAghion @ashkan18 @cavvia

### DIFF
--- a/hokusai/commands/console.py
+++ b/hokusai/commands/console.py
@@ -65,7 +65,7 @@ def console(context, shell, tag, with_config, with_secrets, env):
     job_id = k8s_uuid()
 
   try:
-    call(verbose("kubectl run %s-shell-%s -t -i --image=%s:%s --restart=OnFailure --rm %s -- %s" %
+    call(verbose("kubectl run %s-shell-%s -t -i --image=%s:%s --image-pull-policy=Always --restart=OnFailure --rm %s -- %s" %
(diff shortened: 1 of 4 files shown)


hokusai/commands/console.py
```diff
@@ -65,7 +65,7 @@ def console(context, shell, tag, with_config, with_secrets, env):
     job_id = k8s_uuid()
 
   try:
-    call(verbose("kubectl run %s-shell-%s -t -i --image=%s:%s --restart=OnFailure --rm %s -- %s" %
+    call(verbose("kubectl run %s-shell-%s -t -i --image=%s:%s --image-pull-policy=Always --restart=OnFailure --rm %s -- %s" %
```
ashkan18: i think by default we have `imagePullPolicy: Always` in our `staging.yml` does this ignore that field?

izakp:
Yes but that needed to be applied to the other commands as well, as they
generate a job spec on-the-fly

bhoggard:
Thanks!

PR #14, opened 2017-03-10
update console and run commands

izakp:
update console and run to inject environment and imagePullPolicy with overrides as json rather than args to kubectl... this should allow working around `kubectl`'s issue parsing env vars with commas in them and allow for injecting env vars such as `FOO=123,456`

cc @joeyAghion @bhoggard @cavvia


ashkan18:
💯  thanks for fixing this!

PR #16, opened 2017-04-05
V1 refactor

izakp:
Refactor for a v1-beta release:
- Remove dependency on `aws` cli - use boto3 python library
- Change `AWS_REGION` env var to `AWS_DEFAULT_REGION` as it is what boto3 expects
- Remove `config` command and drop references to config maps - all env should be managed through the `secrets` subcommand
- Change `init` command to `setup`
- Add example `secretKeyRef` fields to service dependencies (i.e. `REDIS_URL`) in `setup`
- Refactor exception handling & exit codes to be handled top-level by `hokusai.command` decorator
- Remove `console` command - add a `--tty` option to `run` command instead
- Remove `pull` command so as not to pollute local registry
- Change `images` command to `tags` command, which lists remote image tags via ECR API
- `push` command always builds project (if the `test` command previously built the container, docker will use the same image cache)
- Remove `build` command - builds are implicit in `test` and `push`
- Add `--skip-build` option to `dev` command
- Add `logs` command
- Refactor internal APIs - add Kubectl, Deployment, Service classes
- Change `stack` subcommands to `create` / `update` / `delete` / `status`
- Support for multiple deployments via `deployments` array in config.yml
- Update `setup` command to create ECR repo if it does not exist
- Update `check` command to check ECR repository existence and kubectl contexts 'staging' and 'production'
- Add basic unittests

Breaking changes:
- Existing projects should move all environment vars to secretKeyRefs, rather than configMapKeyRefs before updating.


hokusai/commands/stack.py
```diff
@@ -1,72 +1,72 @@
...
+  shout(kctl.command("delete -f %s" % kubernetes_yml))
   print_green("Stack %s deleted" % kubernetes_yml)
-  return 0
 
+@command
 def stack_status(context):
-  HokusaiConfig().check()
+  deployment = Deployment(context)
+  deployment_data = []
+  for item in deployment.cache:
```
bhoggard: What are `deployment.cache` and `service.cache` used for?

README.md
```diff
@@ -17,21 +17,17 @@ If you use homebrew, install Docker for Mac with: `brew tap caskroom/cask && bre
...
 2) [Docker Compose](https://docs.docker.com/compose/)
 
-If you installed Docker for Mac, `docker-compose` is also installed. Otherwise install with: `pip install docker-compose`.
-
-3) [AWS CLI](http://docs.aws.amazon.com/cli/latest/userguide/installing.html)
-
-Install with: `pip install awscli`. Set the `$AWS_ACCESS_KEY_ID` and `$AWS_SECRET_ACCESS_KEY` and `$AWS_REGION` environment variables. You should have permissions to evaluate the `aws ecr get-login` comamnd for your ECR region and push access to your ECR repositories.
+If you installed Docker for Mac, `docker-compose` is also installed. Otherwise install with: `(sudo) pip install docker-compose`.
 
 ## Setup
```
bhoggard: Should this be called Installation now, since setup is a different thing, for apps?

README.md
```diff
@@ -50,14 +46,13 @@ Required options:
...
   - `--s3-key`: The S3 key of your org's kubectl config file
 
-### Initializing a project
+### Setting up an existing project
 
-* `hokusai init` - Writes hokusai project config to `hokusai/config.yml`, creates test, development and production yaml files alongside it, and adds a Dockerfile to the current directory.
+* `hokusai setup` - Writes hokusai project config to `hokusai/config.yml`, creates test, development and production yml files alongside it, and adds a Dockerfile to the current directory.
 
 Required options:
   - `--aws-account-id`: Your AWS account ID - can be found in your AWS account console.
```
bhoggard: Should we add the other options - mongodb, etc. so people realize they are here without running `--help`?

hokusai/commands/stack.py
izakp: To avoid making the same API calls (i.e. it loads the data from kubect when it is instantiated, and accesses it in different methods)

README.md
```diff
@@ -83,23 +76,29 @@ Recommended approach is to upload your `kubectl` config to S3 and use following
 ```bash
 hokusai install --s3-bucket <bucket name> --s3-key <file key>
 ```
-### Working with ConfigMaps
 
-* `hokusai config pull` - Pulls config from the Kubernetes server and writes to the `hokusai` directory.
-* `hokusai config push` - Pushes config from the hokusai directory to the Kubernetes server. Config is created for the project as the Kubernetes ConfigMap object `{project}-config`
+The following commands refer to a Kubernetes context.  By convention, Hokusai looks for both a `staging` and a `production` context available to `kubectl` (usually in `~/.kube/config`).
```
izakp: Ok will expand the docs in README to include arguments / options

README.md
```diff
@@ -50,14 +46,13 @@ Required options:
   - `--s3-bucket`: The S3 bucket containing your org's kubectl config file
   - `--s3-key`: The S3 key of your org's kubectl config file
 
-### Initializing a project
+### Setting up an existing project
 
-* `hokusai init` - Writes hokusai project config to `hokusai/config.yml`, creates test, development and production yaml files alongside it, and adds a Dockerfile to the current directory.
+* `hokusai setup` - Writes hokusai project config to `hokusai/config.yml`, creates test, development and production yml files alongside it, adds a Dockerfile to the current directory, and creates a project ECR repo.
```
cavvia: isn't it actually called 'yaml'? seems more human readable than the file extension.

README.md
```diff
@@ -83,23 +76,29 @@ Recommended approach is to upload your `kubectl` config to S3 and use following
...
+
 ### Working with Stacks
 
-* `hokusai stack up` - Launch a stack for a given Kubernetes context.
-* `hokusai stack down` - Delete a stack defined for a given Kubernetes context.
-* `hokusai stack status` - Print the stack status.
+* `hokusai stack create` - Create a stack in the given Kubernetes context.
+* `hokusai stack update` - Update a stack in the given Kubernetes context.
+* `hokusai stack delete` - Delete a stack in the given Kubernetes context.
+* `hokusai stack status` - Print the stack status in the given Kubernetes context.
```
cavvia: i prefer these to `up` and `down`, more descriptive. :+1:

README.md
```diff
@@ -108,5 +107,14 @@ hokusai install --s3-bucket <bucket name> --s3-key <file key>
 
 ### Running a console
 
-* `hokusai console` - Launch a container and attach a shell session.
 * `hokusai run` - Launch a container and run a given command. It exits with the status code of the command run in the container (useful for `rake` tasks, etc).
```
cavvia: +1 I want to know how to launch a console not just a command.

hokusai/deployment.py
```diff
@@ -0,0 +1,70 @@
+import datetime
+import json
+
+from hokusai.config import config
+from hokusai.kubectl import Kubectl
+from hokusai.common import print_red, print_green, shout, get_ecr_login
+
+class Deployment(object):
```
cavvia: i would go for more of a command object pattern here, so instead of `update` the method would be `run` or something like that.

hokusai/kubectl.py
```diff
@@ -0,0 +1,25 @@
+import json
+
+import yaml
+
+from hokusai.common import shout
+
+class Kubectl(object):
```
cavvia: It feels a bit weird for all these classes to live in the same namespace (Kubectl, ECR, Deployment) as they don't have a great deal in common.

cavvia:
Phew lots going on here. Hokusai now wraps `kubectl`, `ecr` and `docker` commands, so you could split up the class namespaces a bit perhaps. Still a bit light on tests but since it's mainly I/O i'm not sure how you could improve on that.

bin/hokusai
```diff
@@ -163,18 +119,17 @@ def secrets(action, context, secret, verbose):
...
-  if action == 'up':
-    sys.exit(hokusai.stack_up(context))
-  elif action == 'down':
-    sys.exit(hokusai.stack_down(context))
+  if action == 'create':
+    hokusai.stack_create(context)
+  elif action == 'update':
+    hokusai.stack_update(context)
+  elif action == 'delete':
+    hokusai.stack_delete(context)
```
cavvia: isn't this called `stack_destroy` below?

hokusai/commands/stack.py
```diff
@@ -1,72 +1,72 @@
...
-  print_green("Stack %s created" % kubernetes_yml)
-  return 0
+  kctl = Kubectl(context)
+  shout(kctl.command("apply -f %s" % kubernetes_yml))
+  print_green("Stack %s updated" % context)
 
-def stack_down(context):
-  HokusaiConfig().check()
+@command
+def stack_destroy(context):
```
cavvia: this won't destroy you local config file for the given context - is that clear to the end user?

README.md
izakp: YAML is itself a [recursive acronym](https://en.wikipedia.org/wiki/YAML) so we're playing with fire here ;)  Hokusai is looking for `*.yml` files, perhaps it should look for `*.yaml` as well.

hokusai/kubectl.py
izakp: Should they live in `libs` perhaps?

bin/hokusai
izakp: Yea that bit me in my Lunch and Learn :( A good test case.

hokusai/deployment.py
izakp: Yea that might be an appropriate pattern... `run` / `execute`

izakp:
@cavvia thanks for the notes - still adding tests (they are definitely not sufficient yet)

README.md
izakp: Thanks @bhoggard good point - Install Hokusai, Install Dependencies (`deps` as we determined yesterday) and `setup` for apps

README.md
```diff
@@ -69,10 +64,8 @@ Required options:
 
 ### Working with Images
 
-* `hokusai build` - Build the latest docker image for the project.
-* `hokusai pull` - Pull images for your project from your AWS ECR repo.
-* `hokusai push` - Push a locally built image to your AWS ECR repo.
-* `hokusai images` - List all project images in your local docker registry.
+* `hokusai push` - Build and push and image to the AWS ECR project repo.
```
izakp: Good catch

izakp:
Pushed changes:

- Changes `install` command to `deps` - Install dependencies
- Fixes to Secrets after going through hokusai_demo - empty Secrets object is now created and destroyed along with the stack
- Changed kubectl {CONTEXT} argument to a required option wherever applicable.  This does away with confusion about the order of arguments in the `stack` and `secrets` commands.  i.e. you can write `hokusai stack status --context staging` or `hokusai stack --context staging status`
- Changed `hokusai promote` to have default options `--from=staging` and `--to=production`

bhoggard:
Thanks for the changes, @izakp. They all make sense to me.

hokusai/commands/setup.py
```diff
@@ -0,0 +1,239 @@
...
+from jinja2 import Environment, PackageLoader
+env = Environment(loader=PackageLoader('hokusai', 'templates'))
+
+from hokusai.command import command
+from hokusai.config import config
+from hokusai.ecr import ECR
+from hokusai.common import print_green, print_red, build_service, build_deployment, YAML_HEADER
+
+@command
+def setup(aws_account_id, project_type, project_name, aws_ecr_region, port,
```
cavvia: this function is about 200 lines too long :)

hokusai/kubectl.py
cavvia: kubectl and ECR wrappers could go in a `services` module.

test/integration/test_ecr.py
```diff
@@ -0,0 +1,44 @@
...
+
+  @httpretty.activate
+  def test_registry(self):
+    httpretty.register_uri(httpretty.POST, "https://ecr.us-east-1.amazonaws.com/",
+                            body=open(os.path.join(os.getcwd(), 'test', 'fixtures', 'ecr-repositories-response.json')).read(),
+                            content_type="application/x-amz-json-1.1")
+    repositories = [str(repo['repositoryName']) for repo in self.ecr.registry()]
+    self.assertTrue('foo' in repositories)
+    self.assertTrue('bar' in repositories)
+    self.assertTrue('baz' in repositories)
```
cavvia: this testing style with service responses as fixtures looks good to me.

test/integration/test_ecr.py
```diff
@@ -0,0 +1,44 @@
...
+  def test_project_repository_exists(self):
+    httpretty.register_uri(httpretty.POST, "https://ecr.us-east-1.amazonaws.com/",
+                            body=open(os.path.join(os.getcwd(), 'test', 'fixtures', 'ecr-repositories-response.json')).read(),
+                            content_type="application/x-amz-json-1.1")
+    self.assertTrue(self.ecr.project_repository_exists())
+
+  @httpretty.activate
+  def test_create_project_repository(self):
+    httpretty.register_uri(httpretty.POST, "https://ecr.us-east-1.amazonaws.com/",
+                            body=open(os.path.join(os.getcwd(), 'test', 'fixtures', 'ecr-repositories-response.json')).read(),
```
cavvia: shouldn't this be the `create-respositories-response`? I would assert for existence of the `createdAt` property in the response, it's too generic currently.

cavvia:
This is now almost two weeks old and sufficiently large that it's difficult to review - I would look to merge as soon as possible. Perhaps with the introduction of the new integration tests this is good to go - do you have more tests you're planning to roll into this?

A more test driven approach will serve us well in future.

Will the breaking changes require a migration for our existing hokusai apps, and is there some way of codifying that migration? I would be curious to see how you go about migrating to `secretKeyRefs` - does it involve manually redefining every secret for each project as a separate hokusai/kubectl command?

hokusai/commands/setup.py
izakp: Yeh agreed :/

test/integration/test_ecr.py
izakp: 👍

izakp:
@cavvia yea it's a monster - I was just thinking that myself.  I want to commit integration tests against the Kubernetes API but I as of yet cannot get minikube to run in a CI environment.  So going forward,  what I want to do is publish Hokusai to pip early (where master is now) as a 0.1 beta release.  Then we can install it from pip in all existing apps, and work on moving them over to the v 1.0 release as needed.  This will be a breaking migration for existing apps as it will necessitate moving all environment vars over to secrets, but will also address automatic injection of secrets as env vars so kubernetes yaml files do not have to define `secretKeyRef` for every secret key / value pair.

README.md
```diff
@@ -69,44 +68,94 @@ Required options:
...
 ```bash
-hokusai install --s3-bucket <bucket name> --s3-key <file key>
+hokusai deps --s3-bucket <bucket name> --s3-key <file key>
 ```
-### Working with ConfigMaps
-
-* `hokusai config pull` - Pulls config from the Kubernetes server and writes to the `hokusai` directory.
-* `hokusai config push` - Pushes config from the hokusai directory to the Kubernetes server. Config is created for the project as the Kubernetes ConfigMap object `{project}-config`
 
+When running `hokusai setup` `staging.yml` and `production.yml` are created in the hokusai project directory. These files define what Hokusai calls "stacks", and Hokusai is opinionated about a workflow between a staging and production Kubernetes context.  Hokusai commands such as `stack`, `secrets`, `deploy` and `logs` require invocation with either the `--staging` or `--production` flag, which references the respective stack YAML file and interacts with the respective Kubernetes context.
```
joeyAghion: Having `--staging` and `--production` as flags themselves rather than options (like `-c staging` or something) seems arbitrary to me. We should do whatever's typical in the pip/cli world I guess.

bin/hokusai
```diff
@@ -6,232 +6,219 @@ import sys
...
-  set - set {SECRET} in the kubectl {CONTEXT} - each {SECRET} must be in of form 'KEY=VALUE'\n
-  unset - Unset {SECRET} from the kubectl {CONTEXT} - each {SECRET} must be of the form 'KEY'
-  """
-  set_output(verbose)
-  if action == 'get':
-    sys.exit(hokusai.get_secrets(context))
+def secrets(action, secret, staging, production, verbose):
+  """Actions: [get|set|unset]
+
+  create - create the secret Kubernetes object for project
```
joeyAghion: Maybe this should just be automatic if it doesn't exist (when `set` is called)?

joeyAghion:
These changes address a lot of my early confusions. 

The one part that I find unexpected as a developer/consumer is everything being stored as secrets. I like how the cluster is the authoritative reference for all config values (rather than local or committed files), but treating them all 1 way seems blunt. (Said another way, if I were choosing just 1 of them to build first, it might be the more general-purpose configmaps... assuming they could be get/set the same way.)

izakp:
Ok merging for a v1-beta-1 release candidate

PR #17, opened 2017-05-19
Prepare v1.0 beta release

izakp:
Okay so preparing a  v1.0b3 release - this includes some of the discussion we have had around storing environment variables in Kubernetes ConfigMap objects rather than Secret objects (eliminating the need to Base64-encode everything) as well as using the directive:
```
envFrom:
- configMapRef:
    name: {project-name}-environment
```
 in the `hokusai run` command.

Other notable changes:
- prints docker output when running `hokusai push` command
- sources version from `./hokusai/VERSION.txt` in `setup.py` (for pip distribution) and adds `hokusai version` command
- rename `hokusai secrets` to `hokusai env` and store env vars in ConfigMap object named `{project-name}-environment`
- rename `hokusai dependencies` to `hokusai configure` requires invocation with `--kubectl--version` matching version of the Kubernetes API server being targeted, as well as `--s3-bucket` and `--s3-key` for location of org's kubectl config file

We will need to migrate existing apps to use the new `envFrom` format described above, similar to what @ashkan18 did in https://github.com/artsy/lewitt-api/pull/213 - but can merge and prepare the new `{project-name}-environment` ConfigMap object using `kubectl`, and `jq`

I think this will be good enough for a beta release!


README.md
```diff
@@ -72,24 +72,26 @@ Required options:
...
+When running `hokusai setup` `staging.yml` and `production.yml` are created in the hokusai project directory. These files define what Hokusai calls "stacks", and Hokusai is opinionated about a workflow between a staging and production Kubernetes context.  Hokusai commands such as `stack`, `env`, `deploy` and `logs` require invocation with either the `--staging` or `--production` flag, which references the respective stack YAML file and interacts with the respective Kubernetes context.
 
-### Working with Secrets
+### Working with Environment Variables
 
-* `hokusai secrets get` - Prints secrets stored on the Kubernetes server
-* `hokusai secrets set` - Sets secrets on the Kubernetes server. Secrets are stored for the project as key-value pairs in the Kubernetes Secret object `{project_name}-secrets`
-* `hokusai secrets unset` - Removes secrets stored on the Kubernetes server
+* `hokusai env create` - Create the Kubernetes configmap object `{project_name}-environment`
```
joeyAghion: Is this required, or will `... env set` create it when necessary?

izakp: That is required

hokusai/commands/run.py
```diff
@@ -39,10 +33,10 @@ def run(context, command, tty, tag, env):
         "containers": [
           {
             "args": command.split(' '),
-            "name": job_name,
+            "name": name,
             "image": image_name,
             "imagePullPolicy": "Always",
-            "env": environment,
+            'envFrom': [{'configMapRef': {'name': "%s-environment" % config.project_name}}],
```
joeyAghion: Does this mean the `.yml` files generated by `hokusai setup` will already include the `envFrom` directive? 👍

README.md
izakp: I think it works better to keep them separate in a workflow, as you should run `hokusai env create --staging` and set env vars before attempting `hokusai stack create --staging`

hokusai/commands/run.py
izakp: Indeed it does!

izakp:
Okay just realized the multiple deployments spec conflicts with the way that Kubernetes links services to deployments: by means of a [label selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) - so I need to create another label on the deployment (currently just using `app`: `{project-name}` to link them together while still reserving `app` for the service selector

joeyAghion:
Sounds OK to me. There _are_ still a lot of similar-sounding terms (`configure`, `setup`, etc.), but I think they make sense individually. One potentially confusing one is that `configure` configures `hokusai` and has nothing to do with configMaps.

izakp:
Yes, I did consider that - but after taking a break over the last weeks and coming back to the project and thinking about naming, I think `deps` is a weird command and `configure` makes sense in that it's configuring your local Hokusai setup... it does require that we use the `env` command and provide a thin layer over the `configMap` Kubernetes API object but I think it will still be clear for anyone who wants to peek under the hood (i.e. using the `-v` flag`)

izakp:
@bhoggard @ashkan18 @cavvia thoughts? opinions?  Do you find this naming convention more clear or more confusing?

izakp:
0e6c2605c8a086019345da3d8ab435937272633f fixes multiple deployment support by adding the `project: {app-name}` label to group all application deployments.   Also adds `hokusai refresh` command to refresh all deployments (useful when updating environment vars - need to refresh to launch new pods

izakp:
Actually, instead of `app` / `project` label selectors, which on second thought I think is misleading, I am going to try grouping all of the deployments / services under the label `app: {project-name}` and introducing a `component` label for the ELB service selector to target

README.md
```diff
@@ -44,7 +44,7 @@ You can add `-v` (Verbose) to most commands which will show you details of the i
 
 ### Installing Dependencies
 
-* `hokusai deps` - installs and configures kubectl
+* `hokusai configure` - installs and configures kubectl
```
cavvia: this seems more standard than `deps` to me :+1:

README.md
```diff
@@ -72,24 +72,26 @@ Required options:
...
+### Working with Environment Variables
 
-* `hokusai secrets get` - Prints secrets stored on the Kubernetes server
-* `hokusai secrets set` - Sets secrets on the Kubernetes server. Secrets are stored for the project as key-value pairs in the Kubernetes Secret object `{project_name}-secrets`
-* `hokusai secrets unset` - Removes secrets stored on the Kubernetes server
+* `hokusai env create` - Create the Kubernetes configmap object `{project_name}-environment`
+* `hokusai env get` - Print envrionment variables stored on the Kubernetes server
+* `hokusai env set` - Set envrionment variables on the Kubernetes server. Environment variables are stored for the project as key-value pairs in the Kubernetes configmap object `{project_name}-environment`
+* `hokusai env unset` - Remove envrionment variables stored on the Kubernetes server
+* `hokusai env delete` - Delete the Kubernetes configmap object `{project_name}-environment`
```
cavvia: the semantics of this are pretty simple and clearer than `secrets` - if i understand correctly, we no longer have to care whether the env var is sensitive or not, so this is nicely unified now.

hokusai/commands/env.py
```diff
@@ -0,0 +1,48 @@
...
+def get_env(context, environment):
+  configmap = ConfigMap(context)
+  configmap.load()
+  if len(environment) == 0:
+    for k, v in configmap.all():
+      print("%s=%s" % (k, v))
+  else:
+    for k, v in configmap.all():
+      if k in environment:
+        print("%s=%s" % (k, v))
```
cavvia: i like that it's easy to view the entire config for an app in one place now.

izakp:
Okay - so updated labels on deployments and services like so:
```
labels:
    app: {project-name}
    layer: {application / database /etc... }
    component: {web / sidekiq / etc...}
```

The pattern applied is to group _all_ stack components with the same `app` label (i.e. application and service components incl. redis / memcached, etc) but further filter the deployments to target via the `layer: application` selector.  Likewise the `{project-name}-web` service targets the `{project-name}-web` deployment via the `component: web` service selector like so:

```
labels:
    app: {project-name}
    layer: application
    component: web
```

This is reflected in the yaml generated by `hokusai setup` as well as in https://github.com/artsy/lewitt-api/pull/216 cc @ashkan18

izakp:
Please update to the newest version on PyPi - `sudo pip uninstall hokusai` / `sudo pip install hokusai==1.0b3`

bin/hokusai
```diff
@@ -173,17 +179,29 @@ def stack(action, staging, production, verbose):
...
   """
   hokusai.lib.common.set_output(verbose)
   context = select_context(staging, production)
   hokusai.deploy(context, tag)
 
 @cli.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--staging', type=click.BOOL, is_flag=True, help='Target staging')
+@click.option('--production', type=click.BOOL, is_flag=True, help='Target production')
+@click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
+def refresh(staging, production, verbose):
```
cavvia: i'm a bit confused as to when one is supposed to use this in the dev workflow - it seems to update the deployment timestamp - isn't this something that should simply happen under the hood during a deployment?

cavvia:
The `env` changes are great, the API looks pretty straightforward to me.

cavvia:
The `app` and `layer` semantics match the opsworks vocabulary so they seem pretty intuitive to me.

izakp:
@joey see https://github.com/artsy/lewitt-api/pull/216

bin/hokusai
izakp: @cavvia so updating the `deploymentTimestamp` is a hack used to trigger a deployment rollout so that it creates a new replicaSet (i.e. spawns new pods) even when the image, or other configuration has not changed.  It should be handled under the hood.

izakp:
I think this is good to merge. Further changes towards v1 will be the addition of more commands / integration tests.

PR #22, opened 2017-06-01
Updates to deploy, stack defaults and dev commands, add history command

izakp:
- Fix bug in retagging images so when redeploying so deployment update does not fail with an ECR error that the tag being deployed already exists for this image manifest

- Template Deployment strategy so that rolling updates ensure HA (always at least one container running)

- Refactor dev command to use actions `start|stop||status|logs|shell|clean` to better work with the docker-compose stack during development workflows

- Add a volume sync to the docker-compose file in `hokusai setup` - this should help local development, even if sync is slow it can be replaced with [docker-sync](http://docker-sync.io/) if a user wants to install and use that rather than docker's own volume syncing

- Add `history` command, printing out the revision history of the project's deployment(s).  I won't implement a rollback command per-se, as when managing multiple deployments, there is the possibility that their revision numbers will go out-of-sync, making relying on a `kubectl rollout undo` [command](http://vishh.github.io/docs/user-guide/kubectl/kubectl_rollout_undo/) worrysome. Let's leave it to devs to use the `history` and `images` commands to see what SHA1 was last deployed, and what they can rollback to.

cc @joeyAghion

### DIFF
--- a/bin/hokusai
+++ b/bin/hokusai
@@ -72,14 +72,42 @@ def check():
   hokusai.check()
 
 @cli.command(context_settings=CONTEXT_SETTINGS)
-@click.option('--skip-build', type=click.BOOL, is_flag=True, help='Do not build the project while launching the stack')
+@click.argument('action', type=click.STRING)
+@click.option('-s', '--skip-build', type=click.BOOL, is_flag=True, help='Do not build the project while launching the stack')
+@click.option('-f', '--follow', type=click.BOOL, is_flag=True, help="Follow stack output when running 'start' or 'logs'")
 @click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
-def dev(skip_build, verbose):
-  """
-  Boot the development stack
-  """
+def dev(action, skip_build, follow, verbose):
+  """Manage the development stack
+
+  Actions: [start|stop||status|logs|shell|clean]
+
+  start - Start the development stack defined in ./hokusai/development.yml
+
+  stop - Stop the development stack defined in ./hokusai/development.yml
+
+  status - Print the status of the development stack
+
+  logs - Print logs from the development stack
+
+  shell - Attach a shell session to the stack's primary project container
+
+  clean - Stop and remove all containers in the stack"""
   hokusai.lib.common.set_output(verbose)
-  hokusai.development(skip_build)
+  if action == 'start':
+    hokusai.dev_start(skip_build, follow)
+  elif action == 'stop':
+    hokusai.dev_stop()
+  elif action == 'status':
+    hokusai.dev_status()
+  elif action == 'logs':
+    hokusai.dev_logs(follow)
+  elif action == 'shell':
+    hokusai.dev_shell()
+  elif action == 'clean':
+    hokusai.dev_clean()
+  else:
+    hokusai.lib.common.print_red("Invalid action %s" % action)
+    sys.exit(-1)
 
 @cli.command(context_settings=CONTEXT_SETTINGS)
 @click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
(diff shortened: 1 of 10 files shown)


bin/hokusai
```diff
@@ -72,14 +72,42 @@ def check():
   hokusai.check()
 
 @cli.command(context_settings=CONTEXT_SETTINGS)
-@click.option('--skip-build', type=click.BOOL, is_flag=True, help='Do not build the project while launching the stack')
+@click.argument('action', type=click.STRING)
+@click.option('-s', '--skip-build', type=click.BOOL, is_flag=True, help='Do not build the project while launching the stack')
+@click.option('-f', '--follow', type=click.BOOL, is_flag=True, help="Follow stack output when running 'start' or 'logs'")
```
cavvia: would `tail` be a more intuitive name for this?

hokusai/lib/common.py
```diff
@@ -69,6 +69,13 @@ def build_deployment(name, image, target_port, layer='application', component='w
     ('metadata', {'name': "%s-%s" % (name, component)}),
     ('spec', {
       'replicas': 1,
+      'strategy': {
+        'rollingUpdate': {
+          'maxSurge': 1,
+          'maxUnavailable': 0
+        },
+        'type': 'RollingUpdate'
+      },
```
cavvia: Great - so no need to define this in projects going forward.

bin/hokusai
izakp: Hmmm yes I like

hokusai/lib/common.py
izakp: Yup yup

hokusai/commands/development.py
```diff
@@ -1,23 +1,68 @@
...
+  shout("docker-compose -f %s exec %s sh" % (docker_compose_yml, config.project_name), print_output=True)
+
+@command
+def dev_clean():
+  docker_compose_yml = os.path.join(os.getcwd(), 'hokusai/development.yml')
+  if not os.path.isfile(docker_compose_yml):
+    print_red("Yaml file %s does not exist." % docker_compose_yml)
+    return -1
+  shout("docker-compose -f %s stop" % docker_compose_yml, print_output=True)
+  shout("docker-compose -f %s rm --force" % docker_compose_yml, print_output=True)
```
cavvia: ok so this command is the one which removes all _stopped_ containers.

izakp: Yep!

cavvia:
Just to clarify the rollback process - I'd run `history`, fetch the sha1 from the last deployment and use the `images` command to check the tag name associated with that sha1, and then run a `deploy` with that tag name?

hokusai/commands/setup.py
```diff
@@ -119,6 +119,7 @@ def setup(aws_account_id, project_type, project_name, aws_ecr_region, port,
       if compose_environment == 'development':
         services[config.project_name]['command'] = development_command
         services[config.project_name]['ports'] = ["%s:%s" % (port, port)]
+        services[config.project_name]['volumes'] = ['../:/app']
```
cavvia: is this the volume sync you refer to? i'm not too clear what this means.

bin/hokusai
izakp: Hmm although `follow` is used in docker-compose as well as kubernetes - also we have this option in `hokusai logs` so it may cause confusion changing it up.  I think it's bes tto stick with the naming applied by the tools Hokusai is wrapping

hokusai/commands/setup.py
cavvia: oh ignore i get this now, had a look at http://docker-sync.io/

bin/hokusai
cavvia: Sure, was just a suggestion.

hokusai/commands/setup.py
izakp: Yes - this means that changes made in the local project directory will be synced to the container's volume mounted at `/app` (where the code lives) - though one would still need to change the dev `command` to a hot-reloading dev server (i.e. shotgun) to see live changes reflected in their dev environment

cavvia:
all looks good - i have a comment about rollback process but that's all.

izakp:
@cavvia yes - you describe the rollback process exactly

cavvia:
Alright, in that case I'll merge this in!

izakp:
@cavvia thanks!  I'll prepare a release :)

PR #23, opened 2017-06-15
Add migration flag to promote and deploy - fix env injection in run command - update production default replicas

izakp:
- Add a `--migration` flag to `deploy` and `promote` commands i.e. `hokusai deploy {SHA1} --migration bundle exec rake db:migrate` - this will launch a container at the tag specified, run the command, and only continue with the deployment / promotion if it succeeds
- Fix injecting env vars to `run` command via the `--env` flag - now verified that it will override any env vars that are present in the existing environment configmap... i.e. if `REDIS_URL=foo` is set prior, via `hokusai env set`, and the command `hokusai run bash --env REDIS_URL=bar` is run, the container will be launched with the environment `REDIS_URL=bar`
- Update `setup` to default to 2 replicas for production deployments

### DIFF
--- a/hokusai/commands/setup.py
+++ b/hokusai/commands/setup.py
@@ -196,9 +196,14 @@ def setup(aws_account_id, project_type, project_name, aws_ecr_region, port,
       if with_rabbitmq:
         environment.append({'name': 'RABBITMQ_URL', 'value': "amqp://%s-rabbitmq/%s" % (config.project_name, urllib.quote_plus('/'))})
 
+      if stack == 'production':
+        replicas = 2
+      else:
+        replicas = 1
+
       deployment_data = build_deployment(config.project_name,
                                           "%s:%s" % (config.aws_ecr_registry, stack),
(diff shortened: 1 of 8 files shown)


hokusai/commands/setup.py
```diff
@@ -196,9 +196,14 @@ def setup(aws_account_id, project_type, project_name, aws_ecr_region, port,
       if with_rabbitmq:
         environment.append({'name': 'RABBITMQ_URL', 'value': "amqp://%s-rabbitmq/%s" % (config.project_name, urllib.quote_plus('/'))})
 
+      if stack == 'production':
+        replicas = 2
```
joeyAghion: Web Scale™️

joeyAghion:
Note that we'd be in a weird state if the migration command fails and stops the deploy. BUT I don't know any way around that. Them's the breaks with state.

cavvia:
Thanks for this!

PR #25, opened 2017-07-31
Remove -e from docker login command

ashkan18:
# Problem
Keep getting following error when trying `hokusai push`

```shell
ERROR: unknown shorthand flag: 'e' in -e
See 'docker login --help'.
```

Looks like `-e` is not an [option](https://docs.docker.com/engine/reference/commandline/login/) anymore.

# Solution
Remove it :) With removing it i was able to push an image to ECR.


izakp:
Thanks @ashkan18

PR #30, opened 2017-09-07
Add `pr_only` option to `hokusai diff`

ashkan18:
# Feature
Current `hokusai diff` shows file diffs of whats going to get promoted, while thats awesome, it's little hard to find what where the commits/PRs.

# Solution
Add ` --pr_only` flag to `hokusai diff` command, using `--pr-only` we will use `git log --merges` instead of `git diff` so we should be able to see what PRs are going live.

### DIFF
--- a/bin/hokusai
+++ b/bin/hokusai
@@ -246,14 +246,15 @@ def promote(migration, verbose):
   hokusai.promote(migration)
 
 @cli.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--pr-only', type=click.BOOL, is_flag=True, help='Show only PR merge commits')
 @click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
-def diff(verbose):
+def diff(pr_only, verbose):
   """
   Print a git diff between the tag currently deployed on production
   and the tag currently deployed on staging
   """
   hokusai.lib.common.set_output(verbose)
-  hokusai.diff()
+  hokusai.diff(pr_only)
 
 @cli.command(context_settings=CONTEXT_SETTINGS)
 @click.argument('command', type=click.STRING)
--- a/docs/Command_Reference.md
+++ b/docs/Command_Reference.md
@@ -79,7 +79,7 @@ Note: Environment variables will be automatically injected into containers creat
 * `hokusai promote` - Update the Kubernetes deployment on production to match the deployment running on staging.
 * `hokusai refresh` - Refresh the project's deployment(s).
 * `hokusai history` - Print the project's deployment(s) history.
-* `hokusai diff` - Print a git diff between the tags deployed on production vs staging
+* `hokusai diff` - Print a git diff between the tags deployed on production vs staging, you can pass `pr_only` to only see PR's merge commits
 
 ### Running a command
 
--- a/hokusai/commands/diff.py
+++ b/hokusai/commands/diff.py
@@ -4,7 +4,7 @@ from hokusai.services.deployment import Deployment
 from hokusai.services.ecr import ECR
 
 @command
-def diff():
+def diff(pr_only=False):
   ecr = ECR()
 
   staging_deployment = Deployment('staging')
@@ -28,4 +28,7 @@ def diff():
       return -1
 
   print_green("Comparing %s to %s" % (production_tag, staging_tag))
-  shout("git diff %s %s" % (production_tag, staging_tag), print_output=True)
+  if pr_only:
+    shout("git log --merges --right-only %s..%s" % (production_tag, staging_tag), print_output=True)
+  else:
+    shout("git diff %s %s" % (production_tag, staging_tag), print_output=True)


hokusai/commands/diff.py
```diff
@@ -28,4 +28,7 @@ def diff():
       return -1
 
   print_green("Comparing %s to %s" % (production_tag, staging_tag))
-  shout("git diff %s %s" % (production_tag, staging_tag), print_output=True)
+  if pr_only:
+    shout("git log --merges --right-only %s..%s" % (production_tag, staging_tag), print_output=True)
```
starsirius: Yay! 👍 💯 🎉  to a PR only option! It may be tricky to get precisely every and only the Github PRs. Some questions I had before when thinking about this:

* Omit merge commits other than the real PR merges. _I think_ this implementation may still include them.
* Handle PRs that were squashed and didn't have a separate PR merge commit.

Maybe we don't need to be perfect here (but we may need to make this option/description clear).

In @jonallured's [deploy_pr](https://github.com/jonallured/deploy_pr) script, he had to do something like [this](https://github.com/jonallured/deploy_pr/blob/cf8e85dca6cd6342f35e20602d233adfe70c3d70/deploy_pr#L42), but there were still missed cases.

starsirius: Also, I feel a clean commit-level diff may be helpful too.

joeyAghion: I agree that a commit-level diff is helpful... I actually assumed that's what `diff` did. And narrowing to just merged PRs is hard when you consider squashes, different forking/upstream practices, etc.

Really `git diff` and `git log` are _such_ different operations that I expect they'll want to support different parameters, too. I'd be in favor of using a completely different hokusai subcommand for the commit-level output, like `hokusai git-log`. Eventually it could maybe support a `--grep` option to help identify just PRs.

ashkan18: oooh interesting diff command in `deploy_pr`. 

I was also thinking of showing all commits and then we can `grep` for only seeing PR commits. With current code it looks like this
```
Changes not staged for commit:
commit 98664694c7f1ac7394c50f6f7d133c82dc100ef2 (upstream/master, master)
Merge: a3968fb 725b658
Author: Chung-Yi Chi <cychi1210@gmail.com>
Date:   Wed Sep 6 13:39:59 2017 -0400

    Merge pull request #266 from ashkan18/post-refund-notification

    Make sure to post notification events when refunded

commit a3968fb08dba34a552fe61a4b291b98075a8e7af
Merge: e293196 1a83704
Author: Ashkan Nasseri <ashkan18@users.noreply.github.com>
Date:   Tue Sep 5 16:12:30 2017 -0400

    Merge pull request #265 from ashkan18/relax-artwork-group-title

    Relax artwork_group.title required flag
```

Now that i've looked at it it actually kind of looks ok, what do you think?

hokusai/commands/diff.py
```diff
@@ -4,7 +4,7 @@
 from hokusai.services.ecr import ECR
 
 @command
-def diff():
+def diff(pr_only=False):
```
izakp: I tend to leave arguments to the `@command` wrapped functions as positional, (not a keyword with a provided default value as here), because this function should only be called as a direct result of command-line invocation and the Click library ensures this value will always be either `True` or `False` when specified as an option with `is_flag=True` as above.

bin/hokusai
```diff
@@ -246,14 +246,15 @@ def promote(migration, verbose):
...
 @click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
-def diff(verbose):
+def diff(pr_only, verbose):
   """
   Print a git diff between the tag currently deployed on production
   and the tag currently deployed on staging
   """
   hokusai.lib.common.set_output(verbose)
-  hokusai.diff()
+  hokusai.diff(pr_only)
```
izakp: As per my comments below, calling `hokusai.diff` with one positional argument `pr_only` is safe, as it will always be provided.

docs/Command_Reference.md
```diff
@@ -79,7 +79,7 @@ Note: Environment variables will be automatically injected into containers creat
 * `hokusai promote` - Update the Kubernetes deployment on production to match the deployment running on staging.
 * `hokusai refresh` - Refresh the project's deployment(s).
 * `hokusai history` - Print the project's deployment(s) history.
-* `hokusai diff` - Print a git diff between the tags deployed on production vs staging
+* `hokusai diff` - Print a git diff between the tags deployed on production vs staging, you can pass `pr_only` to only see PR's merge commits
```
izakp: Here, I would write it as the command-line option format: `--pr-only` as that's how users will invoke it.  (As you may have noticed, the Click library automatically reformats hypens to underscores, and calls the function wrapped by `@cli.command` with the resultant variable name, i.e. `pr_only`, as you have correctly done otherwise :)

hokusai/commands/diff.py
izakp: I'm wondering if assuming the "Git-flow" PR-style pattern is too much of a generalization... I am in favor of showing the commit details with a flag, then grepp'ing for PRs / merge-commits, if that's what the user wants to do.  I would suggest simplifying and dropping the `--merges` flag from the `git log` command, and changing the option name to `--commits` (the internal flag to `commits`). WDYT?

izakp: Or, changing the default `hokusai diff` behavior to show only a `git log` list of commits, then adding mutually exclusive options `--merges` which would invoke, as @ashkan18 you have it now `git log --merges --right-only` and `--full`, which would invoke `git diff`

ashkan18:
finally coming back to this, what do you think to make it part of `promote` command? Basically have something like `hokusai promote --dry-run` or `hokusai promote --diff` which returns all the commits that are going to be promoted?

izakp:
@ashkan18 I like that idea a lot.  `diff` on its own seemed out of place, but that makes it clear how it fits into the workflow.

ashkan18:
closing in favor of #38

PR #38, opened 2017-09-29
Add gitlog, rename diff to gitdiff

ashkan18:
Follow up after #30 

# Change
We want to be able to see commit level diff between production..staging so we can see what's going to be deployed on next `hokusai promote`.

Added new `hokusai gitlog` and also renamed `hokusai diff` to `hokusai gitdiff` so its more descriptive.


izakp:
LGTM!

PR #46, opened 2017-11-20
@anyone Fix pipeline commands in the doc

starsirius:
Looks like in the doc, `gitdiff`, `gitlog`, `promote` are missing the `pipeline` command.


izakp:
Thanks for catching this @starsirius !

PR #48, opened 2018-01-22
allow local config file as an alternative to an s3 bucket/key

saolsen:
allow local config file as an alternative to an s3 bucket/key


izakp:
🎉

PR #49, opened 2018-01-24
WIP: org templates for yml

saolsen:
WIP: do not merge

First step was to convert our current yml generation to use templates. Now can iterate on what values are passed to templates and how they are defined per org. There is some coupling between the yml templates and the Dockerfile templates (which are per project type) We probably want all of that to be org overridable.


izakp:
Looks like we can use the [Python bindings](https://github.com/libgit2/pygit2) to `libgit2` to pull files from a git repo at a given ref.  Authentication is baked-in (should act as git does, and use an SSH keypair if a git repo is supplied with the `git+ssh://` protocol and we can get a copy of a file of a given revision with [git show](http://www.pygit2.org/recipes/git-show.html)

hokusai/commands/setup.py
```diff
@@ -6,17 +6,22 @@
...
 
 @command
-def setup(aws_account_id, project_type, project_name, aws_ecr_region, port, internal):
+def setup(aws_account_id, project_type, project_name, aws_ecr_region, port, internal, template_dir):
+  if template_dir:
+    env = Environment(loader=FileSystemLoader(template_dir))
+  else:
+    env = Environment(loader=PackageLoader('hokusai', 'templates'))
+
+  print("THIS IS THE CORRECT VERSION STILL")
```
izakp: @saolsen lol debug?

saolsen: this is a feature, the software reassures you that everything is fine

PR #54, opened 2018-03-07
add example of `--tty` flag

acjay:
In the course of editing the docs to add the `--tty` flag, it seemed to me like it would be possible to eliminate some redundancy by consolidating the nearly identical `staging` and `production` sections. I'm guessing that duplication was intentional, so totally open to reverting that part. In my mind, while reading it the first time, it was almost kind of distracting in the sense of trying to figure out whether there were important functional distinctions between the `staging` and `production` cases.

### DIFF
--- a/docs/Command_Reference.md
+++ b/docs/Command_Reference.md
@@ -69,54 +69,29 @@ Note: `hokusai staging` `hokusai production` subcommands such as `create`, `upda
   - `hokusai registry push` - Build and push an image to the project registry.
   - `hokusai registry images` - Print image builds and tags in the project registry.
 
+### Working with remotely deployed Kubernetes environments
 
-### Working with the Kubernetes staging environment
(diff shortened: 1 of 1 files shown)


izakp:
Thanks @acjay it's good to get your opinion on this - I had it in the back of my mind when writing duplicate documentation for staging and production.

I think it may be better to write `[staging|production]` instead of `ENVIRONMENT`.

Also we avoided documenting all optional flags for commands (`--tty` included) as you can use `{command} -h / --help` but I think it's worth it to document them all.  I think I should look into auto-documentation for the CLI via https://github.com/click-contrib/sphinx-click

acjay:
Yeah, I think that makes more sense for representing the 2 environment options.

I thought about including more flags, but it occurred to me that opening an interactive session might be a particularly common need, so I thought I'd just highlight that one. I'll add a note about --help though too.

acjay:
BTW, this is super slick. As my first hands-on experience with Hokusai, I like!!

izakp:
Thanks @acjay !  Oooo an interactive session would be even more slick!  Nice idea!

PR #62, opened 2018-04-19
Review apps

ashkan18:
# Problem
We want to be able to easily deploy different images of the same app on k8s.

# Possible Solution
Adding following command:
```
hokusai staging review_app create <name of the app>
```
What this command does is, it copies `staging.yaml`, adds a new `namespace` definition for this app instance and updates the `namespace` in yaml to point to newly created one and stores the result in `<app_name>.yaml`.
The idea is, we provide this `yaml` file and then we need to manually push the image with proper tag and create this deployment.

### DIFF
--- a/hokusai/cli/review_app.py
+++ b/hokusai/cli/review_app.py
@@ -16,8 +16,8 @@ def review_app(context_settings=CONTEXT_SETTINGS):
 @review_app.command(context_settings=CONTEXT_SETTINGS)
 @click.argument('app_name', type=click.STRING)
 @click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
-@click.option('-sf', '--source-file', type=click.STRING, default='hokusai/staging.yml', help='Source deployment file')
-@click.option('-dn', '--destination-namespace', type=click.STRING, default=None, help='Target Namespace')
+@click.option('-sf', '--source-file', type=click.STRING, default='hokusai/staging.yml')
+@click.option('-dn', '--destination-namespace', type=click.STRING, default=None)
 def setup(app_name, verbose, source_file, destination_namespace):
   set_verbosity(verbose)
   # create app_name.yaml file based on source_file
@@ -29,7 +29,7 @@ def setup(app_name, verbose, source_file, destination_namespace):
 @review_app.command(context_settings=CONTEXT_SETTINGS)
 @click.argument('app_name', type=click.STRING)
 @click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
-@click.option('-dn', '--destination-namespace', type=click.STRING, default=None, help='Target Namespace')
+@click.option('-dn', '--destination-namespace', type=click.STRING, default=None)
 def create(app_name, verbose, destination_namespace):
   if destination_namespace is None: destination_namespace = app_name
   destination_namespace = clean_string(destination_namespace)
@@ -38,7 +38,7 @@ def create(app_name, verbose, destination_namespace):
 @review_app.command(context_settings=CONTEXT_SETTINGS)
 @click.argument('app_name', type=click.STRING)
 @click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
-@click.option('-dn', '--destination-namespace', type=click.STRING, default=None, help='Target Namespace')
+@click.option('-dn', '--destination-namespace', type=click.STRING, default=None)
 def env_copy(app_name, verbose, destination_namespace):
   if destination_namespace is None: destination_namespace = app_name
   destination_namespace = clean_string(destination_namespace)
--- a/hokusai/cli/review_app.py
+++ b/hokusai/cli/review_app.py
@@ -6,8 +6,9 @@ import hokusai
 
 from hokusai.cli.base import base
 from hokusai.cli.staging import KUBE_CONTEXT
-from hokusai.lib.common import set_verbosity, CONTEXT_SETTINGS
+from hokusai.lib.common import set_verbosity, CONTEXT_SETTINGS, clean_string
 from hokusai.lib.config import config
+from hokusai.lib.exceptions import HokusaiError
 
 @base.group()
 def review_app(context_settings=CONTEXT_SETTINGS):
@@ -19,20 +20,31 @@ def review_app(context_settings=CONTEXT_SETTINGS):
 @click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
 @click.option('-sf', '--source-file', type=click.STRING, default='hokusai/staging.yml', help='Source deployment file')
 @click.option('-dn', '--destination-namespace', type=click.STRING, default=None, help='Target Namespace')
-def create(app_name, verbose, source_file, destination_namespace):
+def setup(app_name, verbose, source_file, destination_namespace):
   set_verbosity(verbose)
-  # copy staging yml file
+  # create app_name.yaml file based on source_file
   if destination_namespace is None: destination_namespace = app_name
+  destination_namespace = clean_string(destination_namespace)
   create_new_app_yaml(source_file, app_name, destination_namespace)
+
 
 @review_app.command(context_settings=CONTEXT_SETTINGS)
 @click.argument('app_name', type=click.STRING)
 @click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
-def apply(app_name, verbose):
-  """Create the Kubernetes resources defined in ./hokusai/%s.yml""" % app_name
-  set_verbosity(verbose)
+@click.option('-dn', '--destination-namespace', type=click.STRING, default=None, help='Target Namespace')
+def create(app_name, verbose, destination_namespace):
+  if destination_namespace is None: destination_namespace = app_name
+  destination_namespace = clean_string(destination_namespace)
   hokusai.k8s_create(KUBE_CONTEXT, app_name, app_name)
+  hokusai.k8s_copy_config(KUBE_CONTEXT, destination_namespace)
 
+@review_app.command(context_settings=CONTEXT_SETTINGS)
+@click.argument('app_name', type=click.STRING)
+@click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
+def update(app_name, verbose):
+  """Updates the Kubernetes based resources defined in ./hokusai/%s.yml""" % app_name
+  set_verbosity(verbose)
+  hokusai.k8s_update(KUBE_CONTEXT, app_name)
 
 def create_new_app_yaml(source_file, app_name, destination_namespace):
   with open(source_file, 'r') as stream:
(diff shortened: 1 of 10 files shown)


hokusai/cli/review_app.py
```diff
@@ -0,0 +1,35 @@
...
+def create(app_name, verbose):
+  set_verbosity(verbose)
+  # copy staging yml file
+  create_new_app_yaml('hokusai/staging.yml', app_name)
+
+def create_new_app_yaml(original_yaml_file, app_name, current_namespace='default'):
+  with open(original_yaml_file, 'r') as stream:
+    try:
+      yaml_content = list(yaml.load_all(stream))
+      # @TODO: replace namespace: default with new namespace name
```
izakp: I would not make any assumptions that the file configures their app in the `"default"` namespace - it coul dbe any namespace.  So will need to traverse keys and replace any `namespace` keys with the new namespace

joeyAghion:
In general I think `hokusai` commands should wrap the underlying operations as transparently as possible. Why not just add a `--namespace` argument to the existing `create` command rather than a new, triply-nested one?

hokusai/cli/review_app.py
izakp: Maybe need to dig into the Kubernetes object specs a bit more here but AFAIK _all_ objects support the key `metadata.namespace` so it might just be enough to iterate through all top-level objects defined in the source yml and simply overwrite this key

hokusai/cli/review_app.py
```diff
@@ -0,0 +1,35 @@
...
+      # prepend new namespace definition to yaml
+      new_namespace = {"apiVersion": "v1", "kind": "Namespace", "meta": { "name": str(app_name) }}
+      yaml_content = [new_namespace] + yaml_content
+      with open("hokusai/%s.yml" % app_name, 'w') as output:
+        yaml.dump_all(yaml_content, output)
```
izakp: Not sure if this will reorder keys in the yaml serializations of the k8s objects (as dict keys are un-sorted) but this is why we add a [OrderedDict](https://docs.python.org/2/library/collections.html#collections.OrderedDict) Yaml representer https://github.com/artsy/hokusai/blob/master/hokusai/lib/representers.py

hokusai/cli/review_app.py
```diff
@@ -0,0 +1,35 @@
...
+  """Manage/Create review apps"""
+  pass
+
+@review_app.command(context_settings=CONTEXT_SETTINGS)
+@click.argument('app_name', type=click.STRING)
+@click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
+def create(app_name, verbose):
+  set_verbosity(verbose)
+  # copy staging yml file
+  create_new_app_yaml('hokusai/staging.yml', app_name)
```
anandaroop: [Ceiling cat] Seems like if the review app is by definition based on the staging environment, then maybe the proposed cli syntax could drop that detail? So just <tt>hokusai ~staging~ review_app create</tt>?

ashkan18: good point, that actually makes it easier for later if we wanted to do things like `hokusai review_app deploy <app name>`.

ashkan18:
Made some changes in last commit:
- Moved the command to `base` so now it would be
```
hokusai review_apps create <name of app>
```
- Added few new options: 
  - `--source-file`, `-sf`: by default it's `hokusai/staging.yaml` but we now have option to change it
  - `--destination-namespace`, `-dn`: by default it's name of the app but we can override it to what we need

- we now iterate through parsed yaml and we replace all instances of `namespace` to be `desitination_namespace`.

*Question*: Should we actually create the deployment on k8s when we do `hokusai review_apps create`? if we want to do that i need to also update the image to not be `staging` and be `app_name` otherwise i was thinking we make deploy part manual and it's deployer's job to make sure new yaml is properly have image tag.

izakp:
> Question 1: Should we actually create the deployment on k8s when we do hokusai review_apps create?

We can have this command actually create the k8s resources if:
- The new namespace is created _before_ the app itself
- All the review app's resources are within this namespace

Given that we will want to spot-check the new resource config before applying I would tie this to state dependent on a local file / directory structure.  I.e. `hokusai review_apps create` would create the appropriate file and generate a UUID -appended to the filename (PR number?) to then reference via a `hokusai review_apps show` or whatnot

> Question 2: Should we (could we) copy configMap from source namespace to destination?

I think, given the process above - if the steps taken are:

1) create
2) apply

then we can copy the configMap over from the source at apply / launch time...

I would make it a point that we think of review apps as having a lifecycle.

izakp:
> In general I think hokusai commands should wrap the underlying operations as transparently as possible. Why not just add a --namespace argument to the existing create command rather than a new, triply-nested one?

@joeyAghion I don't think passing the `--namespace` argument through to kubectl is viable as what we want to achieve here is a more declarative syntax and to operate across namespaces.  If you `kubectl apply` with the `--namespace` option it will only scope its actions to that particular namespace.

ashkan18:
having trouble with having app with same name on different namespace, getting this error:

```shell
> hokusai review_app apply test_app
Error from server (AlreadyExists): error when creating "/Users/ashkinas/src/volt/hokusai/test_app.yml": deployments.extensions "volt-web" already exists
Error from server (AlreadyExists): error when creating "/Users/ashkinas/src/volt/hokusai/test_app.yml": services "volt-web" already exists
ERROR: Command 'kubectl --context staging create --save-config -f /Users/ashkinas/src/volt/hokusai/test_app.yml' returned non-zero exit status 1
```

hokusai/cli/review_app.py
```diff
@@ -0,0 +1,61 @@
...
+          })
+        ])
+      yaml_content = [new_namespace] + yaml_content
+      with open("hokusai/%s.yml" % app_name, 'w') as output:
+        yaml.safe_dump_all(yaml_content, output, default_flow_style=False)
+    except yaml.YAMLError as exc:
+      print(exc)
+
+def update_namespace(yaml_section, destination_namespace):
+  if 'namespace' in yaml_section: yaml_section['namespace'] = destination_namespace
```
izakp: Would just overwrite namespace here - i.e. drop the `if`

hokusai/cli/review_app.py
```diff
@@ -0,0 +1,61 @@
...
+  if destination_namespace is None: destination_namespace = app_name
+  create_new_app_yaml(source_file, app_name, destination_namespace)
+
+@review_app.command(context_settings=CONTEXT_SETTINGS)
+@click.argument('app_name', type=click.STRING)
+@click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
+def apply(app_name, verbose):
+  """Create the Kubernetes resources defined in ./hokusai/%s.yml""" % app_name
+  set_verbosity(verbose)
+  hokusai.k8s_create(KUBE_CONTEXT, app_name, app_name)
```
izakp: Would use `hokusai.k8s_update` here (also  update the function k8s_update to accept `yaml_file_name` as a keyword argument)- this will then use `kubectl apply` so you can apply updates to the spec

hokusai/cli/review_app.py
```diff
@@ -0,0 +1,61 @@
...
+@review_app.command(context_settings=CONTEXT_SETTINGS)
+@click.argument('app_name', type=click.STRING)
+@click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
+@click.option('-sf', '--source-file', type=click.STRING, default='hokusai/staging.yml', help='Source deployment file')
+@click.option('-dn', '--destination-namespace', type=click.STRING, default=None, help='Target Namespace')
+def create(app_name, verbose, source_file, destination_namespace):
+  set_verbosity(verbose)
+  # copy staging yml file
+  if destination_namespace is None: destination_namespace = app_name
+  create_new_app_yaml(source_file, app_name, destination_namespace)
```
izakp: Perhaps pause here (ask for confirmation?) and then call `hokusai.k8s_create(KUBE_CONTEXT, app_name, app_name)`?

hokusai/cli/review_app.py
```diff
@@ -0,0 +1,61 @@
...
+          ('kind', 'Namespace'),
+          ('metadata', {
+            'name': str(app_name)
+          })
+        ])
+      yaml_content = [new_namespace] + yaml_content
+      with open("hokusai/%s.yml" % app_name, 'w') as output:
+        yaml.safe_dump_all(yaml_content, output, default_flow_style=False)
+    except yaml.YAMLError as exc:
+      print(exc)
```
izakp: `raise HokusaiError("Some error message")` and it will be caught [here](https://github.com/artsy/hokusai/blob/master/hokusai/lib/command.py#L17) or just don't catch the error and it will bubble up and be caught [here](https://github.com/artsy/hokusai/blob/master/hokusai/lib/command.py#L24)

hokusai/cli/review_app.py
ashkan18: but that way this will set `namespace` in each yaml section, we want to only update if that yaml portion has `namespace`, i kind of use this method recursively.

hokusai/cli/review_app.py
ashkan18: oh so you are thinking of this immediately call `create`? i was thinking of this just creating the yaml file and once everything is verified whoever creating this app needs to do `apply` separately.

izakp: That works as well - but whereas `kubectl create` will fail (as you saw) if given resources already exist - `kubectl apply` will update existing resources - so the former is safer when you want to create

hokusai/cli/review_app.py
izakp: Reading more [here](https://github.com/kubernetes/kubernetes/issues/7789) about what happens when a given namespace is defined in the `kubectl` context verses a given yaml spec.  This might be working for us in tests because we are using the `default` namespace in the `context` definition in our `~/.kube/config` file and operations against `yaml` resource specs without defining the namespace work as intended.  In any case it seems this is the behavior implemented:

> If no namespace is specified in the schema, then I'd expect the one from kubeconfig to be used.

> If a namespace is explicitly specified on the command line and that conflicts with the one in the schema, kubectl should raise an error.

This is why I am in favor of putting `namespace` in the schema as it takes precedence over implicit schemas provided by `context`

ashkan18:
Alright made some changes here in command structure, basically to follow commands for staging and production now review apps have:
```
hokusai review_app setup <name>  # creates yaml file
hokusai review_app create <name> # applies yaml file
hokusai review_app update <name> # updates :)
```

ashkan18:
alright, latest command format:
```shell
hokusai review_app setup <name> # creates yaml file
hokusai review_app create <name> # creates namespace/service/deployment, at this stage it will fail till we copy ConfigMap
hokusai review_app copy_env <name> # copies staging ConfigMap from default namespace to new namespace
hokusai review_app update <name> # updates deployment/service/namespace based on yaml
```

hokusai/cli/review_app.py
```diff
@@ -0,0 +1,53 @@
...
+def review_app(context_settings=CONTEXT_SETTINGS):
+  """Manage/Create review apps"""
+  pass
+
+@review_app.command(context_settings=CONTEXT_SETTINGS)
+@click.argument('app_name', type=click.STRING)
+@click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
+@click.option('-sf', '--source-file', type=click.STRING, default='hokusai/staging.yml', help='Source deployment file')
+@click.option('-dn', '--destination-namespace', type=click.STRING, default=None, help='Target Namespace')
+def setup(app_name, verbose, source_file, destination_namespace):
```
izakp: Minor but the click library will [automatically parse](http://click.pocoo.org/5/documentation/) any `@click.command`-wrapped function docstrings and display them as `--help` :)

hokusai/templates/production.yml.j2
```diff
@@ -3,6 +3,7 @@ apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
   name: {{ project_name }}-{{ component }}
+  namespace: default
```
izakp: 👍

izakp:
LGTM!  Will get this merged and into a release!

PR #64, opened 2018-05-03
Release version 0.4.3

izakp:
- `restart` command alias for `refresh` cc @alloy 
- fix ECR auto-discovery with boto to fetch all required ECR project repo metadata (no more required `--aws-account-id` option when running `hokusai setup` can remove all legacy `aws-account-id` / `aws-ecr-region` / `aws-ecr-registry` keys in `./hokusai/config.yml`) cc @ashkan18


alloy:
Nice, thanks! 🙇 

I see you didn’t go the route of educating people to use `refresh` and mentioning `restart` as being deprecated, any reason not to after all?

izakp:
@alloy I documented`restart` as an [alias](https://github.com/artsy/hokusai/blob/master/hokusai/cli/staging.py#L105) for `refresh` - I didn't think it correct to introduce a new command and deprecate it at the same time.

alloy:
Ok.

PR #65, opened 2018-05-04
delete deprecated "local"

acjay:
It looks like the `local` keyword may have been dropped. I think this is accurate now.


izakp:
Oh thank @acjay !  Good catch.

PR #66, opened 2018-05-04
add alternate project types

acjay:
In starting a Hokusai project for Node, I was initially unsure how to apply the directions. I searched in the repo for `ruby-rails` to see if it was obvious what other keywords could be substituted. It was, indeed, obvious, but it seems potentially more obvious to list the project types. The tradeoff would be that this list has to manually be kept in sync. Maybe a more robust solution would be to make the project types list dynamic and inspectable via the CLI? 

Anyway, just thought I'd open a PR in case this seemed like a good idea.

### DIFF
--- a/docs/Getting_Started.md
+++ b/docs/Getting_Started.md
@@ -11,7 +11,7 @@ cd ./path/to/my/rails/project/git/repo
 hokusai setup --project-type ruby-rails
 ```
 
-(Other currently supported project types include `ruby-rack`, `nodejs`, `elixir` and `python-wsgi`.)
+(`hokusai setup --help` can report the list of all supported project types.)
 
 `hokusai setup` will create:
 - A `Dockerfile` in your project root
--- a/docs/Getting_Started.md
+++ b/docs/Getting_Started.md
@@ -11,6 +11,8 @@ cd ./path/to/my/rails/project/git/repo
 hokusai setup --project-type ruby-rails
 ```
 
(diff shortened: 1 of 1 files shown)


izakp:
@acjay the command-line help text is a better source of _all_ available options, so yes it is inspectable via the CLI.

```
$ hokusai setup --help
Usage: hokusai setup [OPTIONS]

  Set up Hokusai for the current project

Options:
  --project-type [ruby-rack|ruby-rails|nodejs|elixir|python-wsgi]
                                  The type of project  [required]
  --project-name TEXT             The project name (default: name of current directory)
  --port INTEGER                  The port of the service (default: 8080)
  --internal                      Create an internal Kubernetes service definition
  --template-dir TEXT             Directory of templates to use.
  -h, --help                      Show this message and exit.
```

izakp:
Is there a way to make it more clear to use the `--help` command when looking for documentation on a specific command?  I put it at the Top of [Command_Reference.md](https://github.com/artsy/hokusai/blob/master/docs/Command_Reference.md#command-reference) but maybe it should be elsewhere.

acjay:
Facepalm! Yeah, didn't try `help`. Just edited to provide that as a suggestion. Let me know what you think now!

izakp:
@acjay thanks!  No worries - it's sometimes just hard to surface the right information in the midst of a _lot_ of information!

PR #67, opened 2018-05-04
fix Yarn failure in CI build

acjay:
I've experience transient issues with `yarn` failing to run due to permission issues when the Dockerfile runs on CircleCI. I feel like this might be a bit of a hack to solve the issue, but it seems to have done the trick.

[Example failied build](https://circleci.com/gh/artsy/volley/12)


izakp:
Strange - had not seen this before but if you don't see the transient issue with this 👍 

Note to self: should separate out our org-specific templates soon so we don't have to depend on a Hokusai release to update them.

Issue #56, opened 2018-03-08
Interactive session


izakp:
Try out https://github.com/click-contrib/click-repl

PR #68, opened 2018-05-07
v0.4.4

izakp:
* 3a45ed0 add interactive 'console' command with click-replj
* 74ce4da aws region fallbacks with boto2 compatibility and default to us-east-1
* aaa241a fix flag collision in logs command -- timestamps now use short flag -s
* 42b3fe7, a034af5, 45628cf update documentation

@acjay this fixes having to set the env var `AWS_DEFAULT_REGION` (and seeing a `No region specified` error) as observed on CI as it will fall back to `us-east-1` if unset in aws / boto credentials or env vars. Also supports the legacy env var `AWS_REGION` as @bhoggard made the point that he has it set in his shells and the boto2 -> boto3 update changed this.

Closes https://github.com/artsy/hokusai/issues/56 also @acjay thanks for the repl idea!  The integration with the Click CLI library via https://github.com/click-contrib/click-repl was easy and is nice to use!

screen shot 2018-05-07 at 7 11 45 pm


hokusai/lib/common.py
```diff
@@ -82,3 +84,13 @@ def k8s_uuid():
...
 def clean_string(str):
   return str.lower().replace('_', '-')
+
+def get_region_name():
+  # boto3 autodiscovery
+  _region = botosession.get_session().get_config_variable('region')
+  if _region:
+    return _region
+  # boto2 compatibility
+  if os.environ.get('AWS_REGION'):
```
izakp: Silly me - `return os.environ.get('AWS_REGION', AWS_DEFAULT_REGION)`

PR #73, opened 2018-07-17
Update review_app docs

anandaroop:
I just had my first adventure with review apps, here are some notes to add to the docs:

- `copy_env` is actually `env_copy`

- `hokusai registry push --force --tag <name>` is probably what you'll need to do

- how to find your public urls

- interim methods for updating env vars and refreshing (issues added for adding subcommands for these: #74, #75)

- some whitespace tweaks.


ashkan18:
awesome! thanks 🙌

PR #77, opened 2018-07-30
Fix using context on k8s_update and not use yaml_file_name

ashkan18:
# Problem
When doing `hokusai review_app update <app name>` we noticed it uses `staging.yml` instead of the review app file yaml.

# Cause
We were setting yaml file based on app name but we were using context (staging for review apps)

# Solution
use `yaml_file_name` instead of context for applying deployment.


izakp:
Thanks @ashkan18 !

PR #79, opened 2018-07-30
More review app doc updates

anandaroop:
Mainly about the crucial step of pointing your review app to the correct image, i.e. the one that you just pushed up to ECR.

Also minor formatting stuff, and a note about `https`.


anandaroop:
I wonder if this whole Review App section should just become its own doc now?

ashkan18:
Nice! 🙌 yeah looks like this part of this doc is taking over, im 👍 with moving it to it's own doc and point from here to it too.

Issue #69, opened 2018-06-13
Setup command templates None as image


izakp:
When running `hokusai setup` it templates `None` as the image repo.  It appears not to be fetching the ECR FQDN correctly.

izakp:
Fixed via https://github.com/artsy/hokusai/pull/80

PR #80, opened 2018-07-31
Prepare version 0.4.6

izakp:
This PR includes:
- Updates for review_apps - support all staging/production commands like `env [get|set|unset|delete]` / `logs` / `deploy` and `refresh`
- Simplify review-app command to always create namespace as `app_name` and use for all subsequent commands
- Add namespace to most service objects - either inherited from context or overridden if passed explicitly
- Always resolve tags in `pipeline` / `deploy` commands to the Git SHA1 tag - ran into issues when getting a `gitlog` / promoting staging when it is deployed to a `staging-{timestamp}` tag
- Add a `--deployment` option to `refresh` so as to only refresh one deployment, if multiple deployments are defined and only refreshing one is desired
- Update output formatters for k8s_status and images. New options for `hokusai [staging|production] status` are:
```
--resources / --no-resources  Print Kubernetes API objects defined in ./hokusai/[staging|production].yml (default: true)
--pods / --no-pods            Print pods (default: true)
--describe                    Print 'kubectl describe' output for resources and pods
--top                         Print top pods
```
- Rename common.yml to build.yml - maintain backwards compatibility
- Bust the registry cache when creating a new repository - fixes https://github.com/artsy/hokusai/issues/69
- add --skip-latest option and logging in push
- Changes `hokusai setup` structure to load templates from a local path or a git remote + branch + path, and only fallback on basic Default configuration.  New options are:
```
--project-name TEXT     The project name (default: name of current directory)
--template-remote TEXT  Git remote of templates to use - you can specify a branch via <git-remote>#<branch>
--template-dir TEXT     Directory of templates to use - can be used with --template-remote
--var TEXT              Extra variables to render Jinja templates in the form of key=value
--allow-missing-vars    Do not fail on undefined template var
```


izakp:
See https://github.com/artsy/artsy-hokusai-templates for setup templates - expands on the awesome work from @saolsen to pull in templates - now they can live in a remote (private) repo and can include our org-specific config - @starsirius @ashkan18 I created a `rails-unicorn` template directory from Volt's current config.  Yet to create `rails-puma` and then a `nodejs` directory.

izakp:
Release candidate available for early adopters :)  @ashkan18 @anandaroop I went through the review_app flow this morning but it could use your 👀 before I publish the release

```
curl https://artsy-provisioning-public/hokusai-v0.4.6 -o /usr/local/bin/hokusai && chmod +x /usr/local/bin/hokusai
```

izakp:
Added `rails-puma` template directory (based on Gravity) and `nodejs` directory (based on Metaphysics) to https://github.com/artsy/artsy-hokusai-templates

izakp:
Merging and releasing

PR #90, opened 2018-11-01
fix error message

bhoggard:
New syntax for pushing images


izakp:
Ooh nice catch - thanks @bhoggard !

PR #91, opened 2018-11-09
Add `hokusai pipeline gitcompare`

ashkan18:
# Problem
Often times we want to be able to get compare link on github so we can post it on our channels let people know what commits are going to be deployed.

# Solution
add `hokusai pipeline gitcompare` which by default uses github as git provider but can get a format string as input for supporting other providers (possibly)

### DIFF
--- a/hokusai/cli/pipeline.py
+++ b/hokusai/cli/pipeline.py
@@ -29,7 +29,7 @@ def gitlog(verbose):
   hokusai.gitlog()
 
 @pipeline.command(context_settings=CONTEXT_SETTINGS)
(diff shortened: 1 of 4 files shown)


hokusai/cli/pipeline.py
```diff
@@ -28,6 +28,16 @@ def gitlog(verbose):
   set_verbosity(verbose)
   hokusai.gitlog()
 
+@pipeline.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--git-compare-link', type=click.STRING, help='Python formatted string input that gets org name, appName and 2 diff sha1s, example: https://github.com/%s/%s/compare/%s...%s', default="https://github.com/%s/%s/compare/%s...%s")
```
izakp: `s/appName/project name`?

PR #92, opened 2018-11-13
Determine if AWS credentials are valid via AWS API call

dblandin:
While Boto3 currently supports [multiple authentication methods][auth methods], we explicitly check that AWS credentials are configured via the environment.

This commit updates the `hokusai check` command to defer an authentication check to Boto3 via the `STS GetCallerIdentity` [API endpoint][].

[auth methods]: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#configuring-credentials
[API endpoint]: https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html

### DIFF
--- a/hokusai/commands/check.py
+++ b/hokusai/commands/check.py
@@ -1,10 +1,13 @@
+import boto3
 import os
+
+import botocore.exceptions as botoexceptions
 
 from hokusai.lib.command import command
 from hokusai.lib.config import config
 from hokusai.services.ecr import ECR
 from hokusai.services.kubectl import Kubectl
-from hokusai.lib.common import print_red, print_green, shout
+from hokusai.lib.common import get_region_name, print_red, print_green, shout
 from hokusai.lib.exceptions import CalledProcessError, HokusaiError
 
 @command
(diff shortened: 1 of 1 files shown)


hokusai/commands/check.py
```diff
@@ -53,16 +56,11 @@ def check_err(check_item):
...
-
-  if os.environ.get('AWS_SECRET_ACCESS_KEY') is not None:
-    check_ok('$AWS_SECRET_ACCESS_KEY')
-  else:
-    check_err('$AWS_SECRET_ACCESS_KEY')
+  try:
+    boto3.client('sts', region_name=get_region_name()).get_caller_identity()
+    check_ok('Valid AWS credentials')
+  except botoexceptions.ClientError, botoexceptions.NoCredentialsError:
+    check_err('Valid AWS credentials')
```
izakp: `s/Valid/Invalid`

dblandin: I was a bit confused by the interface here but I think this expands to an error message of `Valid AWS credentials not found`.

izakp: Oops!  Right you are.

Issue #82, opened 2018-09-19
`hokusai check` fails due to unset kubectl context


javamonn:
I installed and configured `hokusai` according to our internal Artsy docs.

Running `hokusai check` unexpectedly fails on `kubectl not found`, though `kubectl` is installed correctly. Hokusai appears to work correctly even though the check fails. The issue appears to be that the check runs `kubectl version`, which returns a non-zero exit code as `kubectl config current-context` is unset.

The exact output I get from running `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.3", GitCommit:"0480917b552be33e2dba47386e51decb1a211df6", GitTreeState:"clean", BuildDate:"2017-05-10T15:48:59Z", GoVersion:"go1.7.5", Compiler:"gc", Platform:"darwin/amd64"}
Unable to connect to the server: dial tcp: lookup localhost on 8.8.8.8:53: no such host
```

izakp:
Closed via https://github.com/artsy/hokusai/pull/93

PR #93, opened 2018-11-13
Update hokusai check command to find executables via which

dblandin:
For the `hokusai check` command, we want to verify that required executables are available on the system. We currently call an executable's version subcommand, but the invocation varies and it's possible that an implementation may exit nonzero even if the program is available.

Instead of relying on the behavior of various `version` subcommands, simply check that the executable exists in the system `PATH`.

Addresses https://github.com/artsy/hokusai/issues/82

### DIFF
--- a/hokusai/commands/check.py
+++ b/hokusai/commands/check.py
@@ -26,28 +26,28 @@ def check():
     check_err('Config project-name')
 
   try:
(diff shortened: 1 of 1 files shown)


izakp:
LGTM

PR #94, opened 2018-11-13
Report error when kubectl unavailable during context check

dblandin:
During checks of production and staging Kubernetes contexts, report an error when the `kubectl` executable is unavailable.

These checks currently raise an exception when `kubectl` is unavailable:

```
ERROR: Command 'kubectl config view' returned non-zero exit status 127
```

We can instead display errors to the user:

```
staging context not found
production context not found
```

### DIFF
--- a/hokusai/commands/check.py
+++ b/hokusai/commands/check.py
@@ -94,16 +94,20 @@ def check():
     return_code += 1
 
   for context in ['staging', 'production']:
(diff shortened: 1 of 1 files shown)


izakp:
Makes sense that we don't want to error out if `kubectl` is not installed - the check for `kubectl` should fail in this case as well.

Issue #96, opened 2018-11-14
Setup CI builds


izakp:
- Setup test runner on CircleCI
- Run Pyinstaller builds for darwin / linux and publish pre-releases on merges to master
- Flow for creating Github releases synced to pip releases and Docker image builds

PR #98, opened 2018-11-14
Add Circle CI 2.0 configuration

dblandin:
This commit sets up a base Circle CI 2.0 configuration that installs
Python dependencies via Pip and runs the test suite.

It uses the docker executor with the `circleci/python:2` docker image.

Addresses https://github.com/artsy/hokusai/issues/96
Depends on https://github.com/artsy/hokusai/pull/97

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,20 +8,24 @@ jobs:
 
       - restore_cache:
           keys:
-          - v1-dependencies-{{ checksum "requirements.txt" }}
-          - v1-dependencies-
+          - v1-pip-cache-{{ checksum "requirements.txt" }}
+          - v1-pip-cache-
 
       - run:
           name: Install dependencies
           command: |
-            sudo pip install -r requirements.txt
+            mkdir -p ./venv
+            virtualenv ./venv
+            . venv/bin/activate
+            pip install -r requirements.txt
 
       - save_cache:
           paths:
             - ./venv
-          key: v1-dependencies-{{ checksum "requirements.txt" }}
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,27 @@
...
+          - v1-dependencies-
+
+      - run:
+          name: install dependencies
+          command: |
+            sudo pip install -r requirements.txt
+
+      - save_cache:
+          paths:
+            - ./venv
```
izakp: Oooh nice - we already get a virtual env configured and ready to use?

.circleci/config.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      - restore_cache:
+          keys:
+          - v1-dependencies-{{ checksum "requirements.txt" }}
+          - v1-dependencies-
+
+      - run:
+          name: Install dependencies
+          command: |
+            sudo pip install -r requirements.txt
```
izakp: Hmm maybe sudo here will not work.  The `circleci/python:2` image [sets up a `circleci` user](https://github.com/CircleCI-Public/circleci-dockerfiles/blob/master/python/images/2.7.15/Dockerfile#L84) and also [installs `pipenv`](https://github.com/CircleCI-Public/circleci-dockerfiles/blob/master/python/images/2.7.15/Dockerfile#L81) which is a [combination of Pip and Virtualenv](https://pipenv.readthedocs.io/en/latest/)... maybe we should use it as well? Seems better than a `requirements.txt` for a dev/test dependencies while setup.py manages installing dependencies when installing via pip

PR #99, opened 2018-11-14
Fix tests on linux

izakp:
This should fix the failing test "test_command_catches_exceptions" on Linux by not checking for colorized output


dblandin:
CI is green! :tada:

PR #102, opened 2018-11-15
Minor updates

izakp:
- Dev instructions using virtualenvs and editable pip installs
- Minor refactor in configmap.py


dblandin:
Looks like you might need to rebase in order to get that linux spec fix.

PR #104, opened 2018-11-16
Release to PyPi, DockerHub, and GitHub

dblandin:
Add automated version releases to PyPi, DockerHub, and GitHub.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,35 @@ jobs:
       - run: make publish-version
       - run: make publish-latest
 
+  release_pip:
+    docker:
+      - image: python:2
+    steps:
+      - checkout
+      - run: make dependencies
+      - run: make publish-pip
+
+  release_dockerhub:
+    docker:
+      - image: docker:18.09
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run: echo "$DOCKERHUB_PASSWORD" | docker login --username "$DOCKERHUB_USERNAME" --password-stdin
+      - run: apk add make
+      - run: make image
+      - run: make publish-dockerhub
+
+  release_github:
+    docker:
+      - image: golang:1.11
+    steps:
+      - checkout
+      - run: apt-get -qq update
+      - run: apt-get -qq install awscli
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -68,9 +74,30 @@ publish-version:
...
+	twine upload dist/*
+
+publish-dockerhub:
+	docker tag hokusai:latest artsy/hokusai:$(VERSION)
+	docker push artsy/hokusai:$(VERSION)
+	docker tag hokusai:latest artsy/hokusai:latest
+	docker push artsy/hokusai:latest
+
+publish-github:
+	$(AWS) s3 cp \
```
dleve123: This AWS cp copy block is used a couple of other times in this Makefile. Could it make sense / is it possible to extract another "Make function" and use that instead?

dleve123: whoops I misread the order of the args here! carry on

Makefile
```diff
@@ -68,9 +74,35 @@ publish-version:
 	  exit 1; \
 	fi
 
-tag:
-	$(GIT_TAG) --message v$(VERSION) v$(VERSION)
-	$(GIT_PUSH) origin refs/tags/v$(VERSION)
+publish-pip:
+	python setup.py sdist bdist_wheel
+	pip install twine
+	twine upload dist/*
```
izakp: How are PyPi credentials set / injected?  Locally I have `~/.pypirc` set up

.circleci/config.yml
```diff
@@ -63,6 +63,35 @@ jobs:
...
+      - run: make publish-dockerhub
+
+  release_github:
+    docker:
+      - image: golang:1.11
+    steps:
+      - checkout
+      - run: apt-get -qq update
+      - run: apt-get -qq install awscli
+      - run: go get -u github.com/tcnksm/ghr
```
izakp: Nice!

Makefile
dblandin: `TWINE_USERNAME` and `TWINE_PASSWORD` credentials are set in the environment via the Circle CI project configuration.

PR #109, opened 2018-11-20
prepare version 0.4.8

izakp:
Version 0.4.7 release was broken due to https://github.com/artsy/hokusai/commit/9b4bad6617495dd3e6154435321d7c5e2c5a4679 changing the `ConfigMap._to_file()` method but already released.  Removed the faulty released versions (effectively skipping this version now) and provide a bugfix in v0.4.8 via https://github.com/artsy/hokusai/commit/7b57c6845526d651659f6315e379f38e019a4819

Also installing via `pip install .` was broken due to pip attempting to import the hokusai package before installing dependencies.  Fix `setup.py` to read the raw `VERSION` file and change the Docker build to install from source.

This fixes the `make image` task locally and should fix the `release_dockerhub` CI task.  The failed build is here: https://circleci.com/gh/artsy/hokusai/147


izakp:
Good point @dblandin ! Removed that `requires` and changed it to the test steps like other release_ steps

.circleci/config.yml
```diff
@@ -141,7 +141,8 @@ workflows:
               only: release
       - release_dockerhub:
           requires:
-            - release_pip
+            - test_linux
+            - test_macos
```
dblandin: I think you might be a bit behind the upstream master branch. These release jobs no longer require the test jobs as of https://github.com/artsy/hokusai/pull/107.

PR #119, opened 2018-11-28
add openssh client to Docker build

izakp:
Another missing dependency in Alpine... we need the openssh client so git can push with it to remotes

Already hot-pushed new tags


izakp:
cc @dleve123 @ansor4

PR #127, opened 2019-01-02
Prepare version 0.5.1

izakp:
Introduce one new feature in version 0.5.1 - add the optional config key `required-version` to the project's config.yml, and if present check it against the current Hokusai version and bail out if less than the specified required version.

We can use this to guard against running an old version of Hokusai against a project that relies on newer / non-backwards-compatible changes in future releases of Hokusai.


izakp:
@dleve123 changed the config key to `hokusai-required-version` and refactored / added tests and a line in the docs in c785e39

dleve123:
Cool - @izakp merge when ready!

PR #136, opened 2019-02-06
[deploy] Skip git push hooks when pushing tags.

alloy:
⚠️ I made this edit from the GH UI as I didn’t find any tests for this, but please let me know if you want me to test things otherwise.

This was failing deploys for me on a checkout of a repo where I had changes that led to the push hook failing.

I don’t think we ever need to care about push hooks when pushing tags, especially not those for commits that already exist on the remote.

### DIFF
--- a/hokusai/services/deployment.py
+++ b/hokusai/services/deployment.py
@@ -46,7 +46,7 @@ class Deployment(object):
         print_green("Pushing deployment tags to %s..." % remote)
         shout("git tag -f %s" % self.context, print_output=True)
         shout("git tag %s" % deployment_tag, print_output=True)
-        shout("git push --force %s --tags" % remote, print_output=True)
+        shout("git push --force %s --tags --no-verify" % remote, print_output=True)
 
     if config.pre_deploy is not None:
       print_green("Running pre-deploy hook '%s'..." % config.pre_deploy)


hokusai/services/deployment.py
```diff
@@ -46,7 +46,7 @@ def update(self, tag, constraint, git_remote, resolve_tag_sha1=True):
         print_green("Pushing deployment tags to %s..." % remote)
         shout("git tag -f %s" % self.context, print_output=True)
         shout("git tag %s" % deployment_tag, print_output=True)
-        shout("git push --force %s --tags" % remote, print_output=True)
+        shout("git push --force %s --tags --no-verify" % remote, print_output=True)
```
izakp: I was actually thinking to change this to...

```
shout("git push -f --no-verify %s refs/tags/%s" % (remote, self.context), print_output=True)
shout("git push --no-verify %s refs/tags/%s" % (remote, deployment_tag), print_output=True)
```
... so as to only make sure to only push these tags, rather than any others you may have lying around locally... wdyt?

alloy: Not sure what the `self.context` change is about, but a I assume you mean just the removal of `-f`, right? If so, that makes sense to me 👍

izakp: `self.context` is here either `"staging"` or `"production"`, so this is pushing this pointer tag as well as the deployment timestamp tag only, rather than `git push --tags` which will push all local tags to the remote

alloy: Ah gotcha. Yeah sounds good 👍

izakp:
Closing in favor of https://github.com/artsy/hokusai/pull/138

PR #139, opened 2019-03-11
create and push git tags individually with --no-verify

izakp:
Since there seems to be some confusion around this, I'll just get this implemented and merged.

cc @alloy @ashfurrow


alloy:
Thanks!

PR #140, opened 2019-03-11
Kubernetes integration tests and deployment rollout updates

izakp:
- Add Kubernetes integration tests
- Refactor deployment rollouts to only push tags after a successful deployment rollout, and automatically roll back failed deployments if cancelled.  Sending a SIGINT (Ctl+C) will cancel a deployment rollout.  Likewise if the deployment specifies `spec.progressDeadlineSeconds` Kubernetes will time out a rollout, rather than stalling indefinitely, and Hokusai will roll _all_ deployments for an application back if any of them time out.
- Small refactors to shelling out and return codes


test/integration/test_kubernetes.py
```diff
@@ -0,0 +1,89 @@
...
+from hokusai.lib.common import shout
+from hokusai.commands import kubernetes
+from hokusai.services.kubectl import Kubectl
+
+
+class TestKubernetes(HokusaiIntegrationTestCase):
+    @classmethod
+    def setUpClass(cls):
+        kubernetes.ECR = MockECR
+        kubernetes.HOKUSAI_CONFIG_DIR = os.path.join(CWD, 'test/fixtures/project/hokusai')
```
izakp: @alloy these integration tests override the `CWD` constant so hokusai believes it is executing from the 'test/fixtures/project/hokusai' directory and use the Yaml fixtures there, hence the consolidation :)

hokusai/services/deployment.py
```diff
@@ -74,12 +58,31 @@ def update(self, tag, constraint, git_remote, resolve_tag_sha1=True):
...
 
-    print_green("Waiting for rollout to complete...")
+    print_green("Waiting for deployment rollouts to complete...")
 
     rollout_commands = [self.kctl.command("rollout status deployment/%s" % deployment['metadata']['name']) for deployment in self.cache]
-    return_code = shout_concurrent(rollout_commands)
-    if return_code:
-      raise HokusaiError("Deployment failed!", return_code=return_code)
+    return_codes = shout_concurrent(rollout_commands, print_output=True)
+    if any(return_codes):
```
izakp: This is where the deployments rollout magic happens - Hokusai is shelling out to `kubectl rollout status {deployment}` for each deployment in the project - if any of these commands fail, it will consider the entire deployment operation failed and move on to rollback by executing `kubectl rollout undo {deployment}` for each deployment. `kubectl rollout status` will return `1` if the deployment's `progressDeadlineSeconds` is exceeded.  For `apiVersion: extensions/v1beta1` type deployments, no `progressDeadlineSeconds` is set by default (hence the stalled deployments we have been seeing.  For `apiVersion: apps/v1` type deployments it is set to a default of `600` seconds.  So we should set this field everywhere or upgrade our deployment objects to  `apiVersion: apps/v1`. Details in this thread https://github.com/kubernetes/kubernetes/issues/66135

test/integration/test_kubernetes.py
alloy: Aha. In that case I don’t really understand why in some/all cases the `HOKUSAI_CONFIG_DIR` variable still gets joined with `CWD`, e.g. [here](https://github.com/artsy/hokusai/pull/140/files/867da74da910ab7604e190e77d63cbcccbc9db33#diff-5d9ea845219ad1c2d2b2f4dd2098de7eR17). What gives?

hokusai/commands/deployment.py
```diff
@@ -33,9 +33,8 @@ def promote(migration, constraint, git_remote):
   tag = ecr.find_git_sha1_image_tag(tag)
   if tag is None:
     print_red("Could not find a git SHA1 for tag %s.  Aborting." % tag)
-    return -1
+    return 1
```
ashkan18: why switch to 1? this should fail the build right?

hokusai/commands/kubernetes.py
```diff
@@ -14,7 +14,7 @@
 @command()
 def k8s_create(context, tag='latest', namespace=None, yaml_file_name=None):
   if yaml_file_name is None: yaml_file_name = context
-  kubernetes_yml = os.path.join(os.getcwd(), "hokusai/%s.yml" % yaml_file_name)
+  kubernetes_yml = os.path.join(CWD, "%s/%s.yml" % (HOKUSAI_CONFIG_DIR, yaml_file_name))
```
ashkan18: 💯

hokusai/commands/development.py
```diff
@@ -1,14 +1,15 @@
...
+from hokusai import CWD
 from hokusai.lib.command import command
 from hokusai.lib.config import config
 from hokusai.lib.common import print_green, shout, EXIT_SIGNALS
 from hokusai.lib.exceptions import HokusaiError
 
 @command()
 def dev_start(build, detach):
-  docker_compose_yml = os.path.join(os.getcwd(), 'hokusai/development.yml')
+  docker_compose_yml = os.path.join(CWD, 'hokusai/development.yml')
```
ashkan18: should these `hokusai/` change to `HOKUSAI_CONFIG_DIR`?

hokusai/commands/deployment.py
ashkan18: oh is this for https://github.com/artsy/hokusai/pull/140/files#r264610865 ?

hokusai/services/docker.py
```diff
@@ -1,13 +1,15 @@
...
+from hokusai import CWD
 from hokusai.lib.config import config
 from hokusai.lib.common import shout
 from hokusai.lib.exceptions import HokusaiError
 
 class Docker(object):
   def build(self):
-    docker_compose_yml = os.path.join(os.getcwd(), 'hokusai/build.yml')
-    legacy_docker_compose_yml = os.path.join(os.getcwd(), 'hokusai/common.yml')
+    docker_compose_yml = os.path.join(CWD, 'hokusai/build.yml')
```
ashkan18: should these switch to `HOKUSAI_CONFIG_DIR`?

test/integration/test_kubernetes.py
izakp: Ah sorry - the `HOKUSAI_CONFIG_DIR` is overridden there - the `CWD` constant is just to Dry-up calls to `os.getcwd()` across the codebase

alloy: Gotcha, thanks!

.circleci/config.yml
```diff
@@ -9,6 +9,53 @@ jobs:
       - run: make dependencies
       - run: make test
 
+  test_integration:
+    machine:
+      image: circleci/classic:201808-01
```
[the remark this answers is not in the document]
izakp: Hah some "classic" CircleCI machine image? :P

hokusai/commands/deployment.py
izakp: Yeah idk why I put `-1` return code everywhere, which is essentially a `255` return code.  1 makes more sense

hokusai/commands/development.py
izakp: Good call!

hokusai/services/docker.py
izakp: DRY FTW

.circleci/config.yml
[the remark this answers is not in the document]
izakp: Looks like we can build an AMI ourselves as well if we need it https://circleci.com/docs/2.0/vm-service/

hokusai/services/deployment.py
```diff
@@ -120,13 +123,14 @@ def current_tag(self):
 
     for deployment in self.cache:
       containers = deployment['spec']['template']['spec']['containers']
-      container_names = [container['name'] for container in containers]
-      container_images = [container['image'] for container in containers]
+      container_images = [container['image'] for container in containers if self.ecr.project_repo in container['image']]
```
izakp: @ashkan18 here's the fix for the error `Deployment's containers do not reference the same image tag.  Aborting.` you were seeing when running `hokusai pipeline gitcompare`...

izakp:
Going to get this merged and prepare a release!

hokusai/services/deployment.py
ashkan18: perfect! 🙌 THANKS!

PR #142, opened 2019-03-14
Release version 0.5.3

izakp:
## v0.5.3

* d23aafc DRY up references to development, test, build and common yml files
* 867da74 update return value of hokusai/commands/logs.py
* 3295d9f remove the kubectl timeout flag and rely on deployment config to specify progressDeadlineSeconds to orchestrate deployment rollouts (note - this also fixes issues with pipeline commands referencing sidecar containers running alongside application containers)
* 708ffdd create and push git tags individually with --no-verify

cc @ashfurrow @ashkan18 @dblandin @alloy


izakp:
Actually I found a problem with 2c769e37c22c70024250a9c947145ebdcb2eb6a6 which gets subprocess output in a nice way for tests but makes commands like `hokusai test` appear broken as it buffers all output until the subprocess completes....  Going to revert and create a new release

Issue #145, opened 2019-03-18
pre-build hooks not executing when calling hokusai test


izakp:
If a Dockerfile depends on a build arg, and that build arg is set as an environment variable in a `pre-build` hook, it should be available whenever Hokusai calls docker-compose subcommands.

PR #146, opened 2019-03-19
Fix build hooks, Fix git deployment tagging, Add more config options, Refactor post-deploy hooks

izakp:
* 4730592 refactor post-deploy steps to attempt post-deploy hook, pushing Git and ECR tags and print warnings if any fail before exiting 1 - fix CommandRunner.run so config value run-tty is not respected when running migrations, pre-deploy or post-deploy hooks
* 3061785 change order of precedence for configuration - project-specific config takes precedence over environment variable fallbacks
* 97fedd9 fetch from git-remote before creating or pushing tags
* a835e28 update docs for git-remote config option as it does not apply to review apps as well as run-constraints set via an env var
* 2d91f25 fix parsing run-constratins from an env var - accept a comma-delimited string and parse as a list
* 6ea6190 add hokusai.lib.constants and always-verbose to config
* d2b1da4 explicitly tag git deployment tags at the tag to be deployed rather than HEAD
* 2b78d57 add the config options follow-logs, tail-logs and run-constraints
* 76f4ccb call Docker.build() in commands/development.py and commands/test.py rather than passing the --build option to docker-compose so pre- and post-build hooks are executed - fixes https://github.com/artsy/hokusai/issues/145


hokusai/services/deployment.py
```diff
@@ -79,8 +79,8 @@ def update(self, tag, constraint, git_remote, resolve_tag_sha1=True):
       remote = git_remote or config.git_remote
       if remote is not None:
         print_green("Pushing deployment tags to %s..." % remote)
-        shout("git tag -f %s" % self.context, print_output=True)
-        shout("git tag -f %s" % deployment_tag, print_output=True)
+        shout("git tag -f %s %s" % (self.context, tag), print_output=True)
+        shout("git tag -f %s %s" % (deployment_tag, tag), print_output=True)
```
ashfurrow: Yikes, sounds like a tricky one to debug 😄

izakp: Yeah - so this is a pretty gnarly little bug... overlooking passing the target ref as the second argument to `git tag` means that we've been mis-matching tags pushed to git on `hokusai deploy` / `hokusai pipeline promote` commands, as they tagged wherever `HEAD` happened to be rather than the correct tag being deployed :/

ashfurrow: Yikes, yeah. We fixed a very similar bug on our iOS developments just this week: https://github.com/artsy/eigen/pull/2807

docs/Configuration_Options.md
```diff
@@ -26,8 +26,11 @@ When running `hokusai setup` the following files are created:
...
-    - `git-remote`: <string> (optional) - Push deployment tags to git remote when invoking the `hokusai [staging|production] deploy` or the `hokusai pipeline promote` commands.  Can either be a local alias like 'origin' or a URI like `git@github.com:artsy/hokusai.git`.  Bound to the `--git-remote` option for these commands.
-    - `run-tty`: <boolean> (optional) - Attach the terminal to your shell session when invoking `hokusai [staging|production] run`.  Bound to the `--tty` option for this command, overridden by the `HOKUSAI_RUN_TTY` env var.
+    - `git-remote`: <string> (optional) - Push deployment tags to git remote when invoking the `hokusai [staging|production|review_app] deploy` or the `hokusai pipeline promote` commands.  Can either be a local alias like 'origin' or a URI like `git@github.com:artsy/hokusai.git`.  Bound to the `--git-remote` option for these commands.
+    - `run-tty`: <boolean> (optional) - Attach the terminal to your shell session when invoking `hokusai [staging|production|review_app] run`.  Bound to the `--tty` option for this command, overridden by the `HOKUSAI_RUN_TTY` env var.
```
izakp: @ashfurrow I think you brought this up initially, but I now have

```
export HOKUSAI_RUN_TTY=True
export HOKUSAI_TAIL_LOGS=10
export HOKUSAI_FOLLOW_LOGS=True
```

in my `~/.bashrc` so I don't have to type those `--tty` / `--follow` / `--tail` options all the time :)

izakp:
Also including an `always-verbose` / `HOKUSAI_ALWAYS_VERBOSE` config option

izakp:
Seeing some odd build errors... will debug tomorrow

```
python -m unittest discover test.unit
  File "/usr/local/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/local/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/usr/local/lib/python2.7/unittest/__main__.py", line 12, in <module>
    main(module=None)
  File "/usr/local/lib/python2.7/unittest/main.py", line 94, in __init__
    self.parseArgs(argv)
  File "/usr/local/lib/python2.7/unittest/main.py", line 113, in parseArgs
    self._do_discovery(argv[2:])
  File "/usr/local/lib/python2.7/unittest/main.py", line 214, in _do_discovery
    self.test = loader.discover(start_dir, pattern, top_level_dir)
  File "/usr/local/lib/python2.7/unittest/loader.py", line 204, in discover
    raise ImportError('Start directory is not importable: %r' % start_dir)
ImportError: Start directory is not importable: 'test.unit'
Makefile:17: recipe for target 'test' failed
make: *** [test] Error 1
Exited with code 2
```

docs/Configuration_Options.md
izakp: Also including

```
HOKUSAI_RUN_CONSTRAINTS=tier=background
HOKUSAI_ALWAYS_VERBOSE=True
```

ashkan18: nice!

hokusai/commands/logs.py
```diff
@@ -10,10 +10,11 @@ def logs(context, timestamps, follow, tail, namespace=None):
   opts = ''
   if timestamps:
     opts += ' --timestamps'
-  if follow:
+  if follow or config.follow_logs:
     opts += ' --follow'
-  if tail:
-    opts += " --tail=%s" % tail
+  if tail or config.tail_logs:
```
ashkan18: 😍

docs/Configuration_Options.md
joeyAghion: Minor comment: The environment variables shouldn't _override_ the provided flag, but should provide an alternate default. I.e., if the default is usually `false` but the env var is set, the effective default would be `true`. However, passing a flag explicitly (`--tty` in this case) should still take precedence.

Maybe it already works that way and this is just a comment about the docs.

hokusai/services/deployment.py
izakp: One caveat here - this is going to fail after the deployment succeeds, but while creating the git tags if your local git repo does not contain the SHA1 being deployed / promoted...  i.e. you don't have an up-to-date `upstream/master`.

Try and create the deployment tag but do not push it _before_ performing the deploy / promote - push it after a successful rollout.

@alloy @dblandin wdyt, as you ran into problems like this when promoting MP...

izakp: Another option would be to just issue a warning if git tags could not be created / pushed for any reason.

dblandin: We already know the remote via the `--git-remote` flag. Would it be valuable to run a `git fetch` to pull down the sha we're deploying?

docs/Configuration_Options.md
dblandin: I'd also expect the behavior @joeyAghion describes.

Conversely, it would be nice if you could negate a truthy environment default with `--no-tty` or `--no-tail`.

Botocore uses a chain of configuration providers, which looks pretty cool. It would be nice if that was extracted into a library of its own.

>In botocore most of our config chains follow the the precendence: 
> session_instance_variables, environment, config_file, default_value.

https://github.com/boto/botocore/blob/master/botocore/configprovider.py#L125-L126

> Each provider in the chain is called, the first one returning a non-None value is then returned.

https://github.com/boto/botocore/blob/master/botocore/configprovider.py#L295-L296

izakp:
@joeyAghion 

> Minor comment: The environment variables shouldn't override the provided flag, but should provide an alternate default. I.e., if the default is usually false but the env var is set, the effective default would be true. However, passing a flag explicitly (--tty in this case) should still take precedence.

> Maybe it already works that way and this is just a comment about the docs.

This behavior was introduced in https://github.com/artsy/hokusai/pull/101 / v0.4.7

In the case of some of these flags like `--tty` as the default is false (if it is _not_ present) and it's set to true if it _is_ present, its absence can't take precedence over the environment variable.

izakp:
@joeyAghion actually I think the order of precedence you suggest is more intuitive... i.e.
1) Command-line option
2) Config set in `./hokusai/config.yml`
3) Environment variable
4) Default option as provided by the CLI

izakp:
Updated to this behavior in 3061785

```
You can set defaults for certain command options in `./hokusai/config.yml` or in environment variables.  See below for project-specific configuration variables and their environment-variable fallbacks.  The order of precedence for these variables is as follows:

 - If the configuration variable is bound to a command-line option, the option supplied on the command line always takes precedence.
- If the configuration variable is provided in `./hokusai/config.yml` it is used as the option if not supplied on the command line.
- If an environment variable is set it is used as a fallback.
- The default value from the command-line option is used.
```

izakp:
One more commit for this PR - 4730592 

Invoking `CommandRunner.run()` when running migrations, pre-deploy or post-deploy hooks should never fall back to project config or environment config.

Also refactored all post-deploy steps, so once the deployment has been rolled out successfully hokusai will attempt to:
1) Run a post-deploy hook if configured
2) Push ECR tags
3) Push git tags if configured

... such that it will attempt _all_ of these steps before exiting `1`, rather than bailing after the first one that fails.

izakp:
An example of a successful deployment rollout but failure to run a post-deploy hook or push git tags...

```
$ hokusai staging deploy staging
==> hokusai exec `kubectl --context staging get deployment --selector app=delta,layer=application -o json`
Deploying b369d436ff16c90d586b209b0f1cee99f3f50202 to staging...
Patching deployment delta-web...
==> hokusai exec `kubectl --context staging patch deployment delta-web -p '{"spec": {"template": {"spec": {"containers": [{"image": "585031190124.dkr.ecr.us-east-1.amazonaws.com/delta:b369d436ff16c90d586b209b0f1cee99f3f50202", "name": "delta-web"}]}, "metadata": {"labels": {"deploymentTimestamp": "1553096073773976"}}}}}'`
Waiting for deployment rollouts to complete...
==> hokusai exec `kubectl --context staging rollout status deployment/delta-web`
Waiting for rollout to finish: 1 out of 2 new replicas have been updated...
Waiting for rollout to finish: 1 old replicas are pending termination...
Waiting for rollout to finish: 1 old replicas are pending termination...
Waiting for rollout to finish: 1 old replicas are pending termination...
deployment "delta-web" successfully rolled out
Running post-deploy hook 'foo'...
==> hokusai exec `kubectl --context staging run delta-hokusai-run-isacpetruzzi-edhna --attach --image=585031190124.dkr.ecr.us-east-1.amazonaws.com/delta:b369d436ff16c90d586b209b0f1cee99f3f50202 --overrides='{"spec": {"containers": [{"image": "585031190124.dkr.ecr.us-east-1.amazonaws.com/delta:b369d436ff16c90d586b209b0f1cee99f3f50202", "args": ["foo"], "name": "delta-hokusai-run-isacpetruzzi-edhna", "envFrom": [{"configMapRef": {"name": "delta-environment"}}], "imagePullPolicy": "Always"}], "nodeSelector": {"tier": "background"}}, "apiVersion": "v1"}' --restart=Never --rm`
container_linux.go:247: starting container process caused "exec: \"foo\": executable file not found in $PATH"
pod default/delta-hokusai-run-isacpetruzzi-edhna terminated (ContainerCannotRun)
oci runtime error: container_linux.go:247: starting container process caused "exec: \"foo\": executable file not found in $PATH"
WARNING: Running the post-deploy hook failed with return code 127
The deployment 'b369d436ff16c90d586b209b0f1cee99f3f50202' has been rolled out.  However, you should run the post-deploy hook 'foo' manually, or re-run this deployment.
Updating ECR deployment tags in 585031190124.dkr.ecr.us-east-1.amazonaws.com/delta...
Updated ECR tag b369d436ff16c90d586b209b0f1cee99f3f50202 -> staging
Updated ECR tag b369d436ff16c90d586b209b0f1cee99f3f50202 -> staging--2019-03-20--16-35-54
Pushing Git deployment tags to git@github.com:artsy/deltaz.git...
==> hokusai exec `git fetch git@github.com:artsy/deltaz.git`
WARNING: Creating Git deployment tags failed due to the error: 'Command 'git fetch git@github.com:artsy/deltaz.git' returned non-zero exit status 128'
The deployment 'b369d436ff16c90d586b209b0f1cee99f3f50202' has been rolled out.  However, you should create the Git tags 'staging--2019-03-20--16-35-54' and 'staging' manually, or re-run this deployment.
One or more post-deploy steps failed!
```

joeyAghion:
Nice! Happy to help test this.

izakp:
Do folks think it would be a good idea for Hokusai to set a default `progressDeadlineSeconds` for any deployment that fails to specify it?  It would prevent deployment rollouts hanging indefinitely.  See https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#progress-deadline-seconds

izakp:
Going to get this merged and follow up about deployment timeouts

PR #147, opened 2019-03-21
Deployment timeout options, registry images output options, addition logs options and prettier printing

izakp:
* 987a77d (HEAD -> master) add --label option to logs to filter by additional label selectors
* a10105f print newline only after verbose log line
* 1cdb7b9 add the --previous option to logs commands
* 395d57d use a unique tmp dir to download and configure kubectl
* 36316e7 refactor printing always via smart_str and add some spacing
* 538f5d5 add --timeout flag to deploy and promote and patch deployments with `progressDeadlineSeconds` when rolling out
* 8cf1ce4 add flags --reverse-sort, --limit and --filter-tags to hokusai registry images

This PR adds a default timeout of `600` seconds to all deployment rollouts by patching `deployment.spec.progressDeadlineSeconds` when rolling out new deployments.  It can be overridden by the `--timeout` flag on `hokusai [staging|production|review_app] deploy` and `hokusai pipeline promote` commands.  This will prevent deployments from blocking indefinitely if new replicas fail to start or become ready.

Also includes a few extra options for the `hokusai registry images` command... `--reverse-sort` displays images oldest to newest, `--limit` truncates the output of images (default 20) and `--filter-tags` selects only images matching the provided string (so for instance you can use `--filter-tags staging` to select all `staging` `staging-{deployment_timestamp}` tags or even `--filter-tags 2019` to match on tags created this year.  It also colorizes "production" images green and "staging" images yellow for better readability.

Also also includes a few extra options to `hokusai [staging|production] logs` commands... `--previous` retrieves logs from the previous run of a restarted container, so if you see any Pod restarts, get logs right before the container's crash with the `--previous` flag.  You can also pass additional [label selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) to the logs command with the `--label` option ( in the form of `key=value,key=value`) to further filter the Pods from which you are getting logs... so in the case of Gravity, one use the option `--label component=sidekiq` to get only sidekiq pod's logs.

Finally includes a refactor of `print()` statements lying around to use `common.print_smart` which passes through `common.smart_str` and adds some spacing to output.

### DIFF
--- a/docs/Command_Reference.md
+++ b/docs/Command_Reference.md
@@ -115,7 +115,7 @@ When running `hoksuai [staging|production] deploy` or `hokusai pipeline promote`
 2) Attempts to run a `--migration` command, if specified.  If this command fails, Hokusai exits with the return code of this command.
 3) Attempts to run a `pre-deploy` hook, if specified.  If this command fails, Hokusai exits with the return code of this command.
 4) Patches the deployment's containers that run the application image (those that contain the project's repository in their `image` field) with the new deeployment tag.
-5) Waits for the deployment to roll out.  If the deployment does not specify a `progressDeadlineSeconds` it could hang indefinitely.  If the deployment fails to rollout, Hokusai runs a `kubectl rollout undo` to roll all deployments back automatically, then exits with `1`.
+5) Waits for the deployment to roll out.  If the deployment fails to rollout withing the provided `--timeout` (default 10 minutes), Hokusai runs a `kubectl rollout undo` to roll all deployments back automatically, then exits with `1`.
 6) Attempts to run a `post-deploy` hook, if defined.  If it fails, Hokusai prints a warning and continues.
 7) Attempts to push deployment tags to the project registry.  If it fails, Hokusai prints a warning and continues.
 8) Attempts to push deployment tags to Git if `git-remote` is defined in the project's config.  If it fails, Hokusai prints a warning and continues.
--- a/hokusai/cli/pipeline.py
+++ b/hokusai/cli/pipeline.py
@@ -43,10 +43,11 @@ def gitcompare(org_name, git_compare_link, verbose):
 @click.option('--migration', type=click.STRING, help='Run a migration before deploying')
 @click.option('--constraint', type=click.STRING, multiple=True, help='Constrain migration and deploy hooks to run on nodes matching labels in the form of "key=value"')
 @click.option('--git-remote', type=click.STRING, help='Push deployment tags to git remote')
+@click.option('-t', '--timeout', type=click.INT, default=600, help="Timeout deployment rollout after N seconds (default 600)")
 @click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
-def promote(migration, constraint, git_remote, verbose):
+def promote(migration, constraint, git_remote, timeout, verbose):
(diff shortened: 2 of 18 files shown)


dleve123:
Isac - if I’m understanding correctly, this new release would have a default timeout of 10 mins (600 seconds), correct? I would just confirm that all current Artsy deployments are under that threshold. For example, deploying Prediction isn’t fast (approximately 6-7 mins), so it wouldn’t seem out of the question for other services to take longer.

Alternatively, could these patches be released in a 1.X release? If so, that would signal this valuable, but potentially breaking change.

Out of curiosity, what motivated the additional arguments to the `hokusai registry` commands. Was there feedback that there wasn’t a great developer UX there?

izakp:
@dleve123 

> this new release would have a default timeout of 10 mins (600 seconds), correct? I would just confirm that all current Artsy deployments are under that threshold. For example, deploying Prediction isn’t fast (approximately 6-7 mins), so it wouldn’t seem out of the question for other services to take longer.

Yes this would by default time out deployment rollouts after 10 minutes.  This can be adjusted by setting the `--timeout` option so a `hokusai production deploy $SHA --timeout 900` would have a 15 minute timeout.  Since CircleCI generally times out commands after 10 minutes of no output I thought this was a reasonable default.

I don't think we're ready for a 1.0 release, but we could hold off on this change to a `0.6` release, in which case i'd advise projects to pin to a `5.x` release.  I'm not sure it's a breaking change though.

> Out of curiosity, what motivated the additional arguments to the hokusai registry commands. Was there feedback that there wasn’t a great developer UX there?

@alloy provided some feedback asking for more readable output of this command, as well as the sort and filter options

dleve123:
> Yes this would by default time out deployment rollouts after 10 minutes. This can be adjusted by setting the --timeout option so a hokusai production deploy $SHA --timeout 900 would have a 15 minute timeout. Since CircleCI generally times out commands after 10 minutes of no output I thought this was a reasonable default.

Yeah, so presuming that Circle builds use `hokusai test`, which spin up Docker resources, then this timeout isn't really a new constraint on any system. If so, this makes sense to me, and shouldn't be a breaking change (at least not systematically enough to cause alarm). 👍

izakp:
Added a few more commits around `logs` command advanced options and @joeyAghion https://github.com/artsy/hokusai/commit/395d57ddd14772825393d6b7ac1c21d6e3735925 should address those Jenkins job failures we were seeing when running `hokusai configure` in concurrent jobs when downloading `kubectl` - it was a relatively easy fix

izakp:
@dleve123 happy to revert 538f5d5 for this PR and address that in a subsequent release...  Projects can specify the `deployment.spec.progressDealineSeconds` field if they want to be automatically rolled back - this `--timeout` option just means Hokusai takes care of setting that field.

izakp:
Going to merge this and prepare a pre-release for beta testing!

dleve123:
> That said... I'd prefer to release this and start enforcing a timeout on deployment rollouts across all our apps with no config change (only upgrading to a new Hokusai release) - if any rollout takes more than 10 minutes I'd consider it an issue we should try and address, and I'd prefer that deployments are automatically rolled back rather than stalling - i.e. sitting in a half-deployed state, possibly running 2 versions of code, until we take manual intervention.

Agreed!

PR #151, opened 2019-04-05
Removes pip installation from readme

kierangillen:
Removes `pip` installation instructions from readme. As per slack conversation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,16 +35,13 @@ If you installed Docker for Mac, `docker-compose` is also installed. Otherwise i
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


izakp:
Not sure I am totally on-board with this, considering it's a Python project and Pip is the de-facto package manager / installation channel for Python projects.  Also our Pyinstaller builds also support Linux

izakp:
Would instead advise that we publish via homebrew

izakp:
A brew formula should be easy to publish if we just create a .tar.gz archive with the Pyinstaller build in a `./bin/` folder... https://tech.cars.com/distributing-your-home-grown-tools-to-the-masses-with-homebrew-acb7a62518a8

Issue #152, opened 2019-04-10
Verbose mode should redact keys


alloy:
JIRA https://artsyproduct.atlassian.net/browse/PLATFORM-1364

If printing out keys in verbose mode has proven to be a good way to debug things, then I think Hokusai should keep doing so, but redact the keys when printing them out so that we don’t accidentally leak them in CI logs etc. (See https://github.com/artsy/prediction/pull/717 for an example.)

I’m imagining a redaction as commonly applied to credit cards in user-interfaces, i.e. only print the last N characters: `******************************EXAMPLEKEY`

PR #153, opened 2019-04-10
mask verbose output - add a mask when shelling out to docker login

izakp:
Add a keyword argument `mask`, accepting a tuple of 2 regexes to `common.verbose`.  If it is supplied, it unpacks the tuple and passes items as arguments to `re.sub` and only prints the resulting output verbosely.

Also pass the `mask` kwarg through to `verbose` from `returncode`, `shout`, and `shout_concurrent` functions.

When calling `shout(ecr.get_login())` in `hokusai registry push`, mask the username and password.

Fixes https://github.com/artsy/hokusai/issues/152


alloy:
Nice 👌

And such great turnaround time! 💯 / 💯 , would make use of your service again.

Merge on green.

Issue #41, opened 2017-11-15
Pushing local conf via `[staging|production] update` can be hazardous


izakp:
Currently, `hokusai [staging|production] update` pushes config from your local checkout.  As part of a build-and-deploy workflow, it is not guaranteed that anyone running this command will have their local repo in-sync with the latest HEAD on a given branch.

Proposed solution:
- git check that current branch is up-to-date with: `git fetch --all --prune && git diff <branch> <origin/branch>`
- store the config somewhere else, matched to each commit and run an `update` as part of a deploy (this would require refactoring the workflow)

Issue #81, opened 2018-09-06
Fail with informative error message when trying to diff with outdated checkout


alloy:
Instead of:

```
$ hokusai pipeline gitlog
Comparing 2b88e376a9d007df8234b1cd8c17a1838d673d32 to 72a9c0bc4c854cd8e47043651ecbefd56e901f21
fatal: Invalid revision range 2b88e376a9d007df8234b1cd8c17a1838d673d32..72a9c0bc4c854cd8e47043651ecbefd56e901f21
ERROR: Command 'git log --right-only 2b88e376a9d007df8234b1cd8c17a1838d673d32..72a9c0bc4c854cd8e47043651ecbefd56e901f21' returned non-zero exit status 128
```

The tool should tell the user to fetch all latest objects from the remote.

izakp:
Could implement a fetch if `git-remote` is configured for the project -now in https://github.com/artsy/hokusai/pull/101

PR #155, opened 2019-04-24
Introduce Git remote guards 

izakp:
Fetch from git remotes before running `hokusai [staging|production] update` which will enforce a local checkout of master up to date with _all_ remotes unless overridden.

Also perform a `git fetch` for all remotes during `hokusai pipeline git[diff|log]` commands.

Small fixes to integration tests and an error checking for verbose config when running `hokusai setup` before a configfile exists

Fixes https://github.com/artsy/hokusai/issues/41 and https://github.com/artsy/hokusai/issues/81

### DIFF
--- a/hokusai/cli/review_app.py
+++ b/hokusai/cli/review_app.py
@@ -45,7 +45,7 @@ def delete(app_name, verbose):
 def update(app_name, verbose):
   """Updates the Kubernetes based resources defined in ./hokusai/{APP_NAME}.yml"""
   set_verbosity(verbose)
-  hokusai.k8s_update(KUBE_CONTEXT, namespace=clean_string(app_name), yaml_file_name=app_name)
+  hokusai.k8s_update(KUBE_CONTEXT, namespace=clean_string(app_name), yaml_file_name=app_name, skip_checks=True)
 
 
 @review_app.command(context_settings=CONTEXT_SETTINGS)
--- a/hokusai/commands/kubernetes.py
+++ b/hokusai/commands/kubernetes.py
@@ -5,7 +5,7 @@ import yaml
 from hokusai import CWD
 from hokusai.lib.command import command
 from hokusai.lib.config import HOKUSAI_CONFIG_DIR, config
-from hokusai.lib.common import print_green, shout
+from hokusai.lib.common import print_green, shout, returncode
 from hokusai.services.ecr import ECR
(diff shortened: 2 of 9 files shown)


hokusai/commands/kubernetes.py
```diff
@@ -35,12 +35,30 @@ def k8s_create(context, tag='latest', namespace=None, yaml_file_name=None):
...
   if not os.path.isfile(kubernetes_yml):
     raise HokusaiError("Yaml file %s does not exist." % kubernetes_yml)
 
+  current_branch = None
+  for branchname in shout('git branch').splitlines():
+    if '* ' in branchname:
+      current_branch = branchname.replace('* ', '')
+      break
+
+  if not skip_checks:
```
izakp: All the git magic happens here

hokusai/lib/config.py
```diff
@@ -82,15 +82,18 @@ def _env_value_for(self, key, _type):
...
+      with open(HOKUSAI_CONFIG_FILE, 'r') as config_file:
+        config_struct = yaml.safe_load(config_file.read())
+        try:
+          val = config_struct[key]
+        except KeyError:
+          return None
+        if not isinstance(val, _type):
+          raise HokusaiError("Config key %s is not of %s" % (key, _type))
+        return val
+    except IOError:
```
izakp: Bail if config yaml file could not be read for some reason - but this should be guaranteed for all commands except `setup` and `version` by the `config.check()` invocation for [every command by default](https://github.com/artsy/hokusai/blob/master/hokusai/lib/command.py#L16)

hokusai/cli/review_app.py
```diff
@@ -45,7 +45,7 @@ def delete(app_name, verbose):
 def update(app_name, verbose):
   """Updates the Kubernetes based resources defined in ./hokusai/{APP_NAME}.yml"""
   set_verbosity(verbose)
-  hokusai.k8s_update(KUBE_CONTEXT, namespace=clean_string(app_name), yaml_file_name=app_name)
+  hokusai.k8s_update(KUBE_CONTEXT, namespace=clean_string(app_name), yaml_file_name=app_name, skip_checks=True)
```
izakp: Don't make the git checks for review apps

alloy:
I missed this earlier. Nice! 👌

PR #160, opened 2019-06-06
Add registry pull and override Yaml files option

izakp:
@dblandin the registry pull command relates to the work we did last week with Volt and @starsirius 

@ansor4 I hope the addition of the override option `--yaml-file-name` to commands like build, test and push will allow for supporting multiple build targets in a single project...

### DIFF
--- a/hokusai/cli/base.py
+++ b/hokusai/cli/base.py
@@ -47,23 +47,25 @@ def setup(project_name, template_remote, template_dir, var, allow_missing_vars, 
 
 
 @base.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--yaml-file-name', type=click.STRING, help='Use the docker-compose Yaml filename in the ./hokusai directory (default: build.yml)')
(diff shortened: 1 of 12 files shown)


izakp:
> I hope the addition of the override option --yaml-file-name to commands like build, test and push will allow for supporting multiple build targets in a single project...

This would look like (for instance in Vortex)
```
hokusai build --yaml-file-name dbt/build.yml
hokusai test --yaml-file-name dbt/test.yml
```

Where `dbt/test.yml` would source the `vortex` service from `dbt/build.yml` which would reference a `dbt/Dockerfile`...

dblandin:
@izakp What do you think of defining the flag `--file` instead of `--yaml-file-name`?

Would be similar to the `docker-compose` flag:

```
Options:
  -f, --file FILE             Specify an alternate compose file
                              (default: docker-compose.yml)
```

source: https://docs.docker.com/compose/reference/overview/

izakp:
Closing in favor of https://github.com/artsy/hokusai/pull/161

PR #161, opened 2019-06-07
Beta for v0.5.6

izakp:
Follow-up to https://github.com/artsy/hokusai/pull/160

- Add `registry pull` command
- Add the `--filename` option to commands like build, test and push, staging/production create, update, status, delete to specify an override docker-compose / kubernetes Yaml file
- Add a `--service-name` option to test command to get return code of an alternate service
- Add a --tag option to staging/production update commands to roll out an image in concert with a staging/production update in edge cases where this is required cc @eessex @mdole this will allow us to roll out https://github.com/artsy/vibrations/pull/444
- Add a --dry-run flag to staging/production apply command (can use in CI to check config before updating) 
- Create and delete env configmap with staging/production create/delete
- Global config file and updates to configure command so no parameters need be specified if global config is present
- Loosen dependencies and install additional tools aws cli and aws-iam-authenticator in the Docker image


izakp:
Errors in integration tests relating to boto3 updates - will fix Monday

izakp:
Once this is out, we can distribute a global Hokusai config file to `~/.hokusai/config.yml`, in our case:

```
---
kubectl-version: 1.10.7
kubectl-config-file: s3://artsy-citadel/k8s/config-dev
```

Then clients can simply run `hokusai configure` with no parameters

izakp:
Going to get this merged so we get a beta PyInstall version

PR #162, opened 2019-06-10
Prepare version 0.5.6

izakp:
Some additional changes:

Remove --tag from staging/production update command and instead add --update-config and --filename options to staging/production deploy.  This seems to me to make a lot more intuitive sense, as well as in the deployment workflow... rather than patching the existing deployments on the server, it patches the local Yaml file in the same way, then goes through all other deploy steps.

If we want to roll out config with a deployment, all we need to do is:

`hokusai staging deploy TAG --update-config`

Refactoring around the use of the `-f` / `--filename` option - it takes an absolute or relative path 

Only create / delete environment configmap if filename is not overridden in staging/production create/delete.  Avoids deleting the environment configmap if you are using the `--filename` option to say, manage a canary deployment.

Some bugfixes in hokusai/cli/staging.py / hokusai/commands/pull.py / hokusai/services/deployment.py


izakp:
Going to get this out to unblock https://github.com/artsy/positron/pull/2065/ and https://github.com/artsy/vibrations/pull/444

PR #166, opened 2019-07-02
Update env get to sort environment variables before print

ashkan18:
# Problem
When doing 
`hokusai staging env get` the result list is not sorted which makes it hard to follow.

# Solution
Sort it 📉

### DIFF
--- a/hokusai/commands/env.py
+++ b/hokusai/commands/env.py
@@ -16,13 +16,10 @@ def delete_env(context, namespace=None):
 def get_env(context, environment, namespace=None):
   configmap = ConfigMap(context, namespace=namespace)
   configmap.load()
-  if environment:
-    for k, v in configmap.all():
-      if k in environment:
-        print_smart("%s=%s" % (k, v))
-  else:
-    for k, v in configmap.all():
-      print_smart("%s=%s" % (k, v))
+  all_configs = configmap.all()
+  env_vars = {k:v for k,v in all_configs if (environment and k in environment) or not environment }
+  for k in sorted(env_vars.keys()):
+    print_smart("%s=%s" % (k, env_vars[k]))
 
 
 @command()
--- a/hokusai/commands/env.py
+++ b/hokusai/commands/env.py
@@ -17,10 +17,9 @@ def get_env(context, environment, namespace=None):
   configmap = ConfigMap(context, namespace=namespace)
   configmap.load()
   all_configs = configmap.all()
-  env_vars = {k:v for k,v in all_configs if (environment and k in environment) or not environment }
-  for k in sorted(env_vars.keys()):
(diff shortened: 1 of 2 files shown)


hokusai/commands/env.py
```diff
@@ -16,13 +16,10 @@ def delete_env(context, namespace=None):
...
   configmap.load()
-  if environment:
-    for k, v in configmap.all():
-      if k in environment:
-        print_smart("%s=%s" % (k, v))
-  else:
-    for k, v in configmap.all():
-      print_smart("%s=%s" % (k, v))
+  all_configs = configmap.all()
+  env_vars = {k:v for k,v in all_configs if (environment and k in environment) or not environment }
```
ashkan18: @izakp @saolsen is there a better way to do this? `all_configs` is a _generator_ so i couldn't call `sorted` on it, had to first iterate and create the variables and separately sort it.
This is not a big issue unless we have like a million environment variables, but if we have a million env variables we have other problems 😄

saolsen:
🚮

hokusai/commands/env.py
izakp: Haha nice one - Overoptimiztion on my part?  We can easily change the `ConfigMap.all()` method to _not_ be a generator (and if we have a million env vars yes we have other problems) by refactoring its [current implementation](https://github.com/artsy/hokusai/blob/master/hokusai/services/configmap.py#L59) to 
```
return self.struct['data'].items()
```

So we just return a dict and not an iterator 😅 https://stackoverflow.com/questions/10458437/what-is-the-difference-between-dict-items-and-dict-iteritems

izakp: Then you can easily do `sorted(configmap.all().keys())`

ashkan18: good point! did it! 🙌

ashkan18:
@izakp this should be good to go :)

izakp:
@ashkan18 cool - OOO and AFK this weekend but will merge Monday and get it into the next release!

hokusai/commands/env.py
```diff
@@ -16,14 +16,10 @@ def delete_env(context, namespace=None):
...
-  if environment:
-    for k, v in configmap.all():
-      if k in environment:
-        print_smart("%s=%s" % (k, v))
-  else:
-    for k, v in configmap.all():
-      print_smart("%s=%s" % (k, v))
-
+  all_configs = configmap.all()
+  for k in sorted(all_configs.keys()):
```
izakp: More concise (`sorted` will by default to sort by the first element in a tuple):
```
for k, v in sorted(configmap.all().items()):
  if not environment or (environment and k in environment):
    print_smart("%s=%s" % (k, v))
```

izakp:
🎉

PR #168, opened 2019-07-09
Moar features and a deployment tags / ECR module refactor

izakp:
Updates included in this PR to address a few open issues...

- Add the `--update-config` and `--filename` options to the `pipeline promote` command as in the `staging/production deploy` command (forgot to add them in the last release but we've been using them in CI 
- Alias `-u` for `--update-config`
- Add a `--environment` option to `staging/production create` so as to bootstrap a new stack with provided environment variables
- Add a `--configmap` option to `review_app env copy` so we can copy over nginx config when creating a review app and not rely on the kubed daemon to sync configmaps between namespaces that was causing PLEG issues on nodes - see https://artsyproduct.atlassian.net/browse/PLATFORM-1393
- Refactor the ECR module to expose some more useful methods to filter deployment tags and to retrieve the image SHA256 checksum
- Add back the image digests to the `staging/production deploy` command with the `--digests` flag
- Refactor deployments to patch in the image digest rather than the resolving the Git SHA1 tag.  Resolving a tag to another Git SHA1 tag has caused issues where a tag gets orphaned (not attached to an image with another Git SHA1 tag) see https://artsyproduct.atlassian.net/browse/PLATFORM-1604   Furthermore, as we ran into availability issues with ECR, using the image digest [as recommended by Kubernetes](https://kubernetes.io/docs/concepts/configuration/overview/#container-images) allows us to specify the exact image by its SHA256 checksum rather than a tag so Kubernetes will never need to attemt to re-pull a tag

> To make sure the container always uses the same version of the image, you can specify its digest, for example sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2. The digest uniquely identifies a specific version of the image, so it is never updated by Kubernetes unless you change the digest value.

We can then set `imagePullPolicy: IfNotPresent` in our configs to avoid excessive calls to ECR when starting containers

### DIFF
--- a/docs/Review_Apps.md
+++ b/docs/Review_Apps.md
@@ -112,7 +112,13 @@ In order to start a review app you will need to follow these steps:
     hokusai review_app deploy <name> <name>
     ```
 
-14) If you have made changes to your review app's yaml file, you need to update deployment:
+14) If you have made changes to your review app's yaml file, you need to update the deployment:
+
+    ```shell
+    hokusai review_app deploy <name> <name> --update-config
+    ```
+
+    or
 
     ```shell
     hokusai review_app update <name>
--- a/docs/Review_Apps.md
+++ b/docs/Review_Apps.md
@@ -49,13 +49,19 @@ In order to start a review app you will need to follow these steps:
     hokusai review_app create <name>
     ```
 
-6) Copy the staging `configMap` to the new namespace:
+6) Copy the staging environment's `ConfigMap` to the new namespace:
 
     ```shell
     hokusai review_app env copy <name>
     ```
 
-7) Find and visit your staging app:
+7) If necessary, copy other `ConfigMaps` to the new namespace, for example:
+
+    ```shell
+    hokusai review_app env copy <name> --configmap nginx-config
+    ```
+
+8) Find and visit your staging app:
 
     - In the Kubernetes UI, find the "Namespace" dropdown in the main nav and select your chosen `<name>` from that menu
     - Browse to the "Services" section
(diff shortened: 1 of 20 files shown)


docs/Review_Apps.md
```diff
@@ -106,13 +112,13 @@ In order to start a review app you will need to follow these steps:
     hokusai review_app deploy <name> <name>
     ```
 
-13) If you have made changes to your review app's yaml file, you need to update deployment:
+14) If you have made changes to your review app's yaml file, you need to update deployment:
```
eessex: does the `deploy` step accept the new `-u` flag? If so we can probably note that on step 13

izakp: It does!

hokusai/commands/deployment.py
```diff
@@ -25,23 +25,12 @@ def refresh(context, deployment_name, namespace=None):
...
-  tag = ecr.find_git_sha1_image_tag(tag)
-  if tag is None:
-    print_red("Could not find a git SHA1 for tag %s.  Aborting." % tag)
-    return 1
-
+def promote(migration, constraint, git_remote, timeout, update_config=False, filename=None):
   if migration is not None:
     print_green("Running migration '%s' on production..." % migration, newline_after=True)
-    return_code = CommandRunner('production').run(tag, migration, constraint=constraint, tty=False)
+    return_code = CommandRunner('production').run(digest, migration, constraint=constraint, tty=False)
```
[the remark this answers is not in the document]
izakp: Good catch!  That's a bug

izakp:
Going to merge this so we can use a beta build in dev Orbs and Q/A

Issue #174, opened 2019-07-30
env vars can be recast to timestamps


izakp:
ConfigMaps loaded via staging / production env get are parsed, updated then reserialized as Yaml, which has the side effect that timestamp-formatted strings can be parsed into DateTime objects, then researialized in a different format.

Avoid this unnecessary serialization and treat all configmap data as strings

PR #175, opened 2019-08-02
Fix Configmap serialization and other minor issues

izakp:
Addresses https://github.com/artsy/hokusai/issues/174 by serializing / deserializing ConfigMaps to JSON

@starsirius this should cast all new env vars to strings, as they are passed in to the CLI as such

Also few minor fixes relating to Git tags: fix missing HokusaiError imports in `gitcompare.py` and `gitlog.py` and fetch all git tags in `services/deployment.py`


starsirius:
Thanks for helping with this, @izakp! (still catching up on emails from vacation. 😅 )

PR #176, opened 2019-08-14
add --tag-exists option to registry images command

izakp:
To support CI builds that need to determine if an image with a given tag exists

### DIFF
--- a/hokusai/lib/common.py
+++ b/hokusai/lib/common.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 import os
 import sys
 import signal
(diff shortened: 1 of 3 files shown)


hokusai/lib/common.py
```diff
@@ -111,3 +113,9 @@ def get_region_name():
   if os.environ.get('AWS_REGION'):
     return os.environ.get('AWS_REGION')
   return AWS_DEFAULT_REGION
+
+def pick_yes():
+  return random.choice(["Yep", "Si", "Da", "Ja", "Hai", "Jā", "Oui", "Shì", "Sim"])
+
+def pick_no():
+  return random.choice(["Nope", "No", "Nein", "Nay", "Nē", "Non", "Iya", "Méiyǒu", "Não"])
```
dblandin: Haha. Cute :smile:

PR #177, opened 2019-09-13
Use Pyenv and Pipenv to manage hokusai's py version and dependen…

zephraph:
Resolves [PLATFORM-1598](https://artsyproduct.atlassian.net/browse/PLATFORM-1598).

This PR integrates two tools:

- [pyenv](https://github.com/pyenv/pyenv): Like RVM or NVM, used to help manage python versions
- [pipenv](https://github.com/pypa/pipenv): Like npm or bundler, helps manage dependencies

For a more detailed write up of what these are, how they're used, and (most importantly) how they're installed see [this article](https://hackernoon.com/reaching-python-development-nirvana-bb5692adf30c).

## TODO

- [x] Add setup documentation
- [x] Add workflow documentation

PR #179, opened 2019-10-04
Fix build

izakp:
Some minor fixes to the build / release pipeline, following up to https://github.com/artsy/hokusai/pull/177 cc @zephraph @dblandin

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,8 @@ MINOR_VERSION ?= $(shell cat hokusai/VERSION | awk -F"." '{ print $$1"."$$2 }')
 
 dependencies:
 	pip install pipenv --quiet --ignore-installed
-	pipenv install --dev --deploy
+	pipenv install --deploy
+	pip install --quiet pyinstaller==3.3.1
 
 test:
 	pipenv run unit-tests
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -12,7 +12,8 @@ MINOR_VERSION ?= $(shell cat hokusai/VERSION | awk -F"." '{ print $$1"."$$2 }')
 
 dependencies:
 	pip install pipenv --quiet --ignore-installed
-	pipenv install --dev --deploy
+	pipenv install --deploy
+	pip install --quiet pyinstaller==3.3.1
```
zephraph: Could `pyinstaller` just be added as a dev dependency? I'd actually thought I did that..

zephraph: Yeah, it's here: https://github.com/artsy/hokusai/blob/c4828202a54bd11229883c4b9ed368bd9181327a/Pipfile#L7.

zephraph: I see based on the commit message that this was intentional. I'm 👍 with it, but I would like to understand a bit more.

izakp: Pyinstaller is not a dependency per-se - it's a build tool

izakp: So it needs to be callable to create the "binary" distro, but is not part of Hokusai itself

PR #180, opened 2019-10-08
Add template example command

eessex:
Add a template example command and link to templates project.


izakp:
Thanks @eessex - had initially wanted to keep it decoupled from our specific templates but having them as examples do make sense

PR #183, opened 2019-11-21
Futurize - 1st pass at Python 3 compatibility

izakp:
This updates Hokusai to Python 3.  There are a number of idiosyncrasies when upgrading from Python 2 to 3, for example some built-in / standard library functions have different APIs or move to another location.  The [six module](https://six.readthedocs.io/) helps mitigate these issues, and [the future library](https://python-future.org/) tries to help in upgrading and maintaining codebases that run on both versions.  But the tricky part comes with the difference in how Python 2 vs Python 3 [handle unicode versus bytestrings](https://timothybramlett.com/Strings_Bytes_and_Unicode_in_Python_2_and_3.html).

In short, I was able to upgrade to Python 3 and get tests to pass but was not able to make this code work in a backwards-compatible way for Python 2 at first pass.  I'll take another shot at backwards compatibility, but we should increase test coverage and test the PyInstaller distribution as well.

I'd like to keep the futurize branch open and in-sync with master.  Python 2.7 [is officially EOL on January 1st 2020](https://pythonclock.org/), but that doesn't mean it will disappear, just get no further updates.


hokusai/commands/check.py
```diff
@@ -58,7 +58,7 @@ def check_err(check_item):
   try:
     boto3.client('sts', region_name=get_region_name()).get_caller_identity()
     check_ok('Valid AWS credentials')
-  except botoexceptions.ClientError, botoexceptions.NoCredentialsError:
+  except (botoexceptions.ClientError, botoexceptions.NoCredentialsError):
```
izakp: Simple syntax update for exception handling...

hokusai/commands/configure.py
```diff
@@ -5,7 +5,7 @@
 import tempfile
 
 from distutils.dir_util import mkpath
-from urlparse import urlparse
+from six.moves import urllib
```
izakp: An example of a core module moving but API remains the same

hokusai/commands/configure.py
```diff
@@ -36,7 +36,7 @@ def configure(kubectl_version, bucket_name, key_name, config_file, install_to, i
   print_green("Downloading and installing kubectl...", newline_before=True, newline_after=True)
   tmpdir = tempfile.mkdtemp()
   urllib.urlretrieve("https://storage.googleapis.com/kubernetes-release/release/v%s/bin/%s/amd64/kubectl" % (kubectl_version, platform.system().lower()), os.path.join(tmpdir, 'kubectl'))
-  os.chmod(os.path.join(tmpdir, 'kubectl'), 0755)
+  os.chmod(os.path.join(tmpdir, 'kubectl'), 0o755)
```
izakp: More explicit octal notation - backwards compatible

hokusai/lib/common.py
```diff
@@ -36,11 +36,9 @@ def smart_str(s, newline_before=False, newline_after=False):
     s = '\n' + s
   if newline_after:
     s = s + '\n'
-  if isinstance(s, unicode):
```
izakp: @ashkan18 this is where things start to get tricky.  In Python 2, the `str()` constructor is a bytestring, and `unicode` is a unicode string, but in Python 3, `str()` creates a unicode string and `bytes` creates a bytestring.  There are cases (see below) when you want to decode a bytestring to unicode, but in general everything should be unicode.  Do you remember why we made this smart string method and needed the statement...

```
if isinstance(s, unicode):
  return unicode(s).encode("utf-8")
```

I suppose we want to print some special UTF-8 characters so will need to refactor this method

hokusai/services/ecr.py
```diff
@@ -73,7 +73,7 @@ def create_project_repo(self):
 
   def get_login(self):
     res = self.client.get_authorization_token(registryIds=[str(self.aws_account_id)])['authorizationData'][0]
-    token = base64.b64decode(res['authorizationToken'])
+    token = base64.b64decode(res['authorizationToken']).decode()
```
izakp: Gotta decode a bytestring to a string here

Pipfile.lock
```diff
@@ -57,22 +57,6 @@
...
         },
-        "funcsigs": {
-            "hashes": [
-                "sha256:330cc27ccbf7f1e992e69fef78261dc7c6569012cf397db8d3de0234e6c937ca",
-                "sha256:a7bb0f2cf3a3fd1ab2732cb49eba4252c2af4240442415b4abce3b87022a8f50"
-            ],
-            "markers": "python_version < '3.3'",
-            "version": "==1.0.2"
-        },
-        "futures": {
```
izakp: This is a backport of https://docs.python.org/3/library/concurrent.futures.html for 2.7 so unneeded

Pipfile
```diff
@@ -14,7 +14,6 @@ botocore = "==1.12.162"
 click = "==6.7"
 click-repl = "==0.1.6"
 docutils = "==0.14"
-futures = "==3.2.0"
```
izakp: This is a backport of https://docs.python.org/3/library/concurrent.futures.html so unneeded on Python 3 AFAICT

zephraph:
I think it'd be pretty acceptable for us to release a breaking change that no longer supports the 2.x versions. I mean, they _are_ going EOL. I don't know that trying to maintain backwards compatibility is worth the cost.

hokusai/lib/common.py
ashkan18: yeah! i actually don't remember why we had this `smart_str` method, but yes i can imagine we'd need UTF-8 support later.

PR #184, opened 2019-12-04
Remove unneeded command for cleanup steps

pepopowitz:
`hokusai review_app env delete <name>` does not appear to be a valid command. Also, my understanding is that `hokusai review_app delete <name>` will also take care of removing env variables.


izakp:
Thanks @pepopowitz !  Good catch - `hokusai review_app env delete <name>` was an old version's behavior and we since updated Hokusai so that `hokusai review_app delete <name>` will also take care of removing env variables, as you describe.

PR #190, opened 2020-03-31
fea dynamic interpolation #PLATFORM-2249

izakp:
A WIP for dynamic variable interpolation into .yml files.  Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-2249

- Adds ConfigLoader class in `lib/config_loader.py`
- Adds TemplateRenderer class in `lib/template_renderer.py`
- Adds KubernetesSpec in `services/kubernetes_spec.py`
- Render Kubernetes Yaml files via KubernetesSpec when calling `[staging|production|review_app] [create|update|delete|deploy]` and `pipeline promote` commands
- Adds a project config parameter `template_vars_files` and if present / non-empty (this must be a list) then loads template config files in order, merging them before passing to the TemplateRenderer.  This allows us to start with a global org config file, then concatenate / overridde via a local project config file or other shared files.
- Provides project_name and project_repo to templates as default variables so we can reference them in templates  (should this change our setup templates?)

## Future improvements
- Do not render templates on project setup and leave the placeholders for `{{ project_name }}` and `{{ project_repo }}` in place - Note: we should be able to do this without changing code in `hokusai setup` by first updating our templates to use the `{% raw %}` [Jinja tag](https://jinja.palletsprojects.com/en/2.11.x/templates/#escaping) this escape interpolating these variables on setup where we want to, i.e. in production.yml but not in development.yml.
- Hokusai transparently overwrites the image tag on deploy with a SHA256 image checksum - should we instead require that templates contain a reference to `imageTag: {{ deploy_sha }}` instead?  Note that this would require `hokusai [staging|production] update` to check for the currently deployed tag and interpolate it so as not to overwrite the deploy tag...


izakp:
Cool, so @eessex looking into error handling in our TemplateRenderer class, looks like the Jinja API defines how these errors are thrown handled in the template environment https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.Environment

Will attempt to define our own Environment so we can catch and throw validation errrors when it comes to yaml / variable shape

izakp:
Ah nice so according to the [template docs](https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.Template)...

> Normally the template object is generated from an Environment but it also has a constructor that makes it possible to create a template instance directly using the constructor. It takes the same arguments as the environment constructor but it’s not possible to specify a loader.

So we can pass an `undefined` [class](https://jinja.palletsprojects.com/en/2.11.x/api/#undefined-types) in as a parameter to the Template constructor and raise errors if a variable referenced in the template is not part of our vars dictionary

izakp:
I refactored the template config context loaded by ConfigLoader to accept any generic dictionary of key/value pairs.  It is up to the user in templates to reference these as desired, so via `{{ vars.imageTag }}` for example as it is tested.

If we stick to a generic implementation it should be very flexible and we can construct snippets by means of [standard Jinja Template functions](https://jinja.palletsprojects.com/en/2.11.x/templates/).

test/fixtures/project/hokusai/config.yml
```diff
@@ -3,3 +3,5 @@ project-name: hello
 pre-deploy: migrate.sh
 post-deploy: sh -c report.sh
 git-remote: git@github.com:artsy/hokusai.git
+template-config-files:
```
eessex, line 6: I wonder what the use case for this file is-- is it necessary for local dev, is to demonstrate that we can have a local set of vars?

test/fixtures/template_config.yml
```diff
@@ -0,0 +1,4 @@
+---
+vars:
+  containers: [ {'name': 'project-name-web'} ]
```
eessex, line 3: Using `project-name-web` here is confusing to me because it is inconsistent-- in the above `fixtures/kubernetes-config.yml` template does not use the `{{}}` brackets for the metadata fields where this text is used on #5 & 19.

izakp: Ah ok I just added this as an example of the flexibility we have in terms of using a for loop over a list.  What would be a better variable name here?

izakp:
Thanks @eessex - will prepare a release

PR #193, opened 2020-04-09
add twine to publish-pip in Makefile

izakp:
Releasing v0.5.8 [failed](https://circleci.com/gh/artsy/hokusai/833) as twine is only included in dev dependencies so pip install it in the Makefile command

I released v0.5.8 to pip manually

PR #194, opened 2020-04-09
Update Makefile dependencies/build targets

dblandin:
- Install dev dependencies within pyenv
- Run pyinstaller within pyenv

Co-authored-by: @izakp 

I think this might supersede https://github.com/artsy/hokusai/pull/193 ?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,24 @@
 FROM python:2-alpine
 
-# Install Docker
-RUN apk add docker curl
-
-# Install Docker Compose, AWS CLI, pipenv
-RUN pip install docker-compose==1.22.0 && \
-  pip install awscli --upgrade && \
-  pip install pipenv
-
-# Install Git, Ssh, and bash
-RUN apk add git openssh bash
-
-# Install the AWS IAM Authenticator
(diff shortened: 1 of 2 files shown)


izakp:
Thanks @dblandin - I think we still need to `pip install .` in the Dockerfile outside of a Pipenv so it doesn't have to be activated every time the container starts

dblandin:
> I think we still need to pip install . in the Dockerfile outside of a Pipenv so it doesn't have to be activated every time the container starts

@izakp Made a couple more adjustments. Let me know what you think!

Dockerfile
```diff
@@ -1,21 +1,24 @@
...
+RUN curl -L https://github.com/kubernetes-sigs/aws-iam-authenticator/releases/download/v0.4.0/aws-iam-authenticator_0.4.0_linux_amd64 -o /usr/local/bin/aws-iam-authenticator && \
+      chmod a+x /usr/local/bin/aws-iam-authenticator
 
-ADD . /src
-WORKDIR /src
+COPY . /src
 
 # Install Hokusai
-RUN pipenv install --deploy
+RUN pip install .
```
izakp, line 24: 👍

izakp:
Thanks this looks good to me!

izakp:
Darn master build failed - will have a look

PR #195, opened 2020-04-15
Fix markdown

pvinis:
I was curious so I actually tried to run it with the backticks and the dot, and it does start installing, but fails later, whereas normally running it works fine.


izakp:
Haha thanks for catching the typo @pvinis :)

PR #196, opened 2020-04-21
render KubernetesSpec as template on review app setup

izakp:
Fixes errors on `review_app setup` - note that this will render all variables in staging.yml on review app setup and they will not be interpolated in the template on further review app update  /deploys.  Unsure of a workaround as Yaml is invalid until we render the template.  Perhaps this static config is ok for review apps?

@dleve123 once this is merged if you install the `head` build of Hokusai from S3 or use the `head` Docker image tag it should be usable locally or via CI until we cut a release.

### DIFF
--- a/hokusai/commands/namespace.py
+++ b/hokusai/commands/namespace.py
@@ -8,14 +8,19 @@ from hokusai.lib.exceptions import HokusaiError
 from hokusai.lib.common import print_green, clean_string
 from hokusai.lib.constants import YAML_HEADER
 from hokusai.lib.config import HOKUSAI_CONFIG_DIR
+from hokusai.services.kubernetes_spec import KubernetesSpec
 
 @command()
 def create_new_app_yaml(source_file, app_name):
-  with open(source_file, 'r') as stream:
-    try:
-      yaml_content = list(yaml.load_all(stream))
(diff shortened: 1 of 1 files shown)


dleve123:
> Perhaps this static config is ok for review apps?

I think this is fine. Arguably the static nature if better -- particularly for review apps that are geared towards tests k8s/hokusai level config -- less moving pieces.

izakp:
> I think this is fine. Arguably the static nature if better -- particularly for review apps that are geared towards tests k8s/hokusai level config -- less moving pieces.

Okay - I think we should revisit the setup templates and review app setup at some point.  Possibly we don't need to parse the Yaml, inject a namespace object, walk the other objects looking for a namespace key and update it... the `--namespace` option passed to kubectl may be sufficient.  Will have to do some more research.

izakp:
Also yes, Hokusai has the beginnings of an integration test suite (could use some work when it comes to testing against a Kubernetes deployment) - review apps were always overlooked - I will write a test for setting up a review app :)

izakp:
Oh!  Thanks for merging!  Wrote a test for review app create which is odd as it generates the app yaml file into the working directory but anyway figured it out... will follow up with another PR :)

PR #198, opened 2020-04-27
Find all Yaml j2 templates including docker-compose files

izakp:
- Adds TemplateSelector - For a given filename, will look for a file extension yml, yaml, yml.j2 or yaml.j2 in that order, and return the first one it finds so we can experiment with linter settings 
- Remove looking for the "legacy" docker-compose build file which was before named "common"
- Generalize KubernetesSpec to YamlSpec and render docker-compose test / development files as Jinja templates as well (this necessitates Hokusai using a project-local tmp directory)
- Utilize a project-local hokusai tmp directory for other tmpfiles

### DIFF
--- a/hokusai/services/yaml_spec.py
+++ b/hokusai/services/yaml_spec.py
@@ -13,7 +13,7 @@ from hokusai.lib.exceptions import HokusaiError
 
 from hokusai.services.ecr import ECR
 
-class KubernetesSpec(object):
+class YamlSpec(object):
   def __init__(self, kubernetes_yaml):
     self.kubernetes_yaml = kubernetes_yaml
     self.ecr = ECR()
--- a/hokusai/services/yaml_spec.py
+++ b/hokusai/services/yaml_spec.py
@@ -3,9 +3,7 @@ import atexit
 
 import yaml
 
-from tempfile import NamedTemporaryFile
-
-from hokusai.lib.config import config
+from hokusai.lib.config import config, HOKUSAI_TMP_DIR
 from hokusai.lib.config_loader import ConfigLoader
 from hokusai.lib.template_renderer import TemplateRenderer
 
@@ -39,7 +37,7 @@ class YamlSpec(object):
     return rendered_template
 
   def to_file(self):
-    f = NamedTemporaryFile(delete=False)
+    f = open(os.path.join(HOKUSAI_TMP_DIR, os.path.basename(self.kubernetes_yaml)), 'w+b')
     f.write(self.to_string())
     f.close()
     self.tmp_filename = f.name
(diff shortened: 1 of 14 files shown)


hokusai/services/yaml_spec.py
```diff
@@ -1,21 +1,22 @@
...
 from hokusai.lib.template_renderer import TemplateRenderer
 
 from hokusai.lib.exceptions import HokusaiError
 
 from hokusai.services.ecr import ECR
 
-class KubernetesSpec(object):
+class YamlSpec(object):
   def __init__(self, kubernetes_yaml):
     self.kubernetes_yaml = kubernetes_yaml
```
dleve123, line 16: Does it still make sense to name this `kubernetes_yaml`? Should it be something more generic like `yaml`?  Seems like we're generally generalizing the scope of this class, so asking the question

test/integration/test_review_app.py
```diff
@@ -0,0 +1,28 @@
...
+import httpretty
+
+from mock import patch
+
+from test import HokusaiIntegrationTestCase
+
+from hokusai import CWD
+from hokusai.commands.namespace import create_new_app_yaml
+
+class TestReviewApp(HokusaiIntegrationTestCase):
```
dleve123, line 12: 🎊

hokusai/services/yaml_spec.py
izakp: Yeah so now as it would find either Kubernetes or Docker Compose templates and render them as Yaml I chose the generic?

hokusai/services/yaml_spec.py
```diff
@@ -1,21 +1,22 @@
...
     self.ecr = ECR()
+    self.tmp_filename = None
+    atexit.register(self.cleanup)
```
izakp, line 19: Magic here to clean up tmp file handles when the interpreter exits... @dleve123 you mentioned wanting to take a look at the rendered Yaml files - we could add a flag to leave them in place?

hokusai/services/yaml_spec.py
dleve123: Not totally following your comment, but just thinking if it would make to just renaming this from `kubernetes_yaml` to `yaml` - more of a question for you than a suggestion, have pretty little context on this specific codebase.

hokusai/services/yaml_spec.py
dleve123: oh yeah, do think it would be nice to see the final rendered template in some debug mode

izakp:
Though I was a little nervous that these changes would be overengineering (treating docker-compose files as templates the same as kubernetes files ) I think it makes sense?   And we can then figure out how we want to handle with file extensions / linters.

izakp:
Merging to get a `head` build

hokusai/services/yaml_spec.py
izakp: Cool so now if `DEBUG` env var is set hokusai will leave the rendered templates in `.hokusai-tmp` rather than printing to stderr

izakp: See https://github.com/artsy/hokusai/pull/199 for a couple of tweaks for template debugging

PR #205, opened 2020-06-02
Follow "extends" key while rendering docker-compose templates to other templates

izakp:
Parse the docker-compose test or development file after it is rendered for any "extends" key references and also render them into the .hokusai-tmp directory as templates.

This fixes the following error:

With the release of Hokusai v0.5.9 all docker-compose Yaml files are treated as Jinja2 templates, and rendered into the .hokusai-tmp directory before being passed to docker-compose.  A side-effect of this is that references to other docker-compose files in the hokusai directory by means of a relative path to the build.yml file via the extends configuration are sometimes broken.

This error does not manifest when running hokusai test and hokusai dev start as these commands by default render the build.yml file to the .hokusai-tmp directory then run the command docker-compose -f build.yml -p hokusai build before running docker-compose -f test.yml -p hokusai up or docker-compose -f development.yml -p hokusai up respectively, unless the --no-build option is set, in which case the build step is skipped and the command fails with the error:

ERROR: .FileNotFoundError: [Errno 2] No such file or directory: '/Users/isacpetruzzi/src/artsy/convection/.hokusai-tmp/build.yml'
  File "hokusai/lib/command.py", line 17, in wrapper
  File "hokusai/commands/development.py", line 92, in dev_run
  File "hokusai/lib/common.py", line 78, in shout
  File "subprocess.py", line 186, in check_call
CalledProcessError: Command 'docker-compose -f /Users/isacpetruzzi/src/artsy/convection/.hokusai-tmp/development.yml -p hokusai run convection bash' returned non-zero exit status 1

This error also effects the commands hokusai dev [clean|logs|run|status|stop]

Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-2383


hokusai/lib/docker_compose_helpers.py
```diff
@@ -0,0 +1,25 @@
...
+from hokusai.lib.template_selector import TemplateSelector
+from hokusai.services.yaml_spec import YamlSpec
+
+def follow_extends(docker_compose_yml):
+  with open(docker_compose_yml, 'r') as f:
+    rendered_templates = []
+    try:
+      struct = yaml.safe_load(f.read())
+      for service_name, service_spec in struct['services'].iteritems():
+        extended_filename = service_spec['extends']['file']
```
izakp, line 17: Perhaps this isn't great implicit control flow - this line will raise a `KeyError` if the spec does not contain an extends key and the `except` block below will swallow it

izakp: Fixed in 32177c5ccc46b7c10cac18b443d97bdc3e82ce70

PR #212, opened 2020-06-30
python 3 compatibility

izakp:
Updates for python-3 compatibility - addresses https://artsyproduct.atlassian.net/browse/PLATFORM-2545

The code remains backwards-compatible to Python 2.7.16 - wherever possible I adopted shared methods like `dict.items()` which exists in Python 2 and 3 rather than `dict.iteritems()` which exists only in Python 2.

In some cases classes, functions or modules were moved while retaining the same API such as `StringIO` or `urlparse`  in such cases I opted to check the interpreter version and make conditional imports... i.e.
```
if sys.version_info[0] >= 3:
  # import from python 3 
else:
  # import from python 2
```

Also did some cleanup of dependencies (though did not make any core dependency updates) and refactored tests where appropriate.

See https://python-future.org/compatible_idioms.html for a quick overview of the differences in Python 2 / 3 idioms.

# TODO

- [x] Get tests running against python 3
- [x] Fallback to Python 2.7 for import modules where appropriate
- [x] Get PyInstaller build working for Python 3

I also opted to test on both Python 2 and 3 on Linux, but the Darwin Python 3 build time using a brew-installed pyenv and a pyenv-installed Python 3.5.8 takes 14 minutes currently.  See comments below for details.

We can merge and get a Python 3.5.8-built Pyinstaller distribution out as beta - any isntallations via pip on Python 2 should continue working due to backwards-compatibility and we can update at our convenience.


.circleci/config.yml
```diff
@@ -92,34 +92,44 @@ jobs:
       xcode: "9.0"
     steps:
       - checkout
-      - run: make dependencies
-      - run: make test
+      - run: brew update
```
izakp, line 95: Not keen on this but without updating the pyenv formula we can't install python 3.5.8

hokusai/lib/common.py
```diff
@@ -36,9 +36,13 @@ def smart_str(s, newline_before=False, newline_after=False):
     s = '\n' + s
   if newline_after:
     s = s + '\n'
-  if isinstance(s, unicode):
-      return unicode(s).encode("utf-8")
-  elif isinstance(s, int) or isinstance(s, float):
+  if sys.version_info[0] >= 3:
```
izakp, line 39: Wonder if this should be memoized...

pyproject.toml
```diff
@@ -5,13 +5,11 @@ description = "Hokusai is a Docker + Kubernetes CLI for application developers."
 authors = ["Isac Petruzzi <isac.petruzzi@gmail.com>"]
 
 [tool.poetry.dependencies]
-python = "~=2.7"
 boto3 = "==1.9.162"
 botocore = "==1.12.162"
 click = "==6.7"
 click-repl = "==0.1.6"
 docutils = "==0.14"
-futures = "==3.2.0"
```
izakp, line 14: Not sure how this ended up here - not used in the project

.circleci/config.yml
```diff
@@ -89,28 +97,38 @@ jobs:
 
   test_macos:
     macos:
-      xcode: "9.0"
+      xcode: "11.6.0"
     steps:
       - checkout
+      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install pyenv
```
izakp, line 103: Tried the "xcode 12" [executor](https://circle-macos-docs.s3.amazonaws.com/image-manifest/v3358/index.html) the contains system Python 3.8.5 (pip 20.1.1) but tests failed with SSL errors - it's not compiled against openssl libs correctly so have to go with a pyenv-installed Python

Running `HOMEBREW_NO_AUTO_UPDATE=1 brew install pyenv` takes 14 minutes :/ https://circleci.com/gh/artsy/hokusai/1156

izakp:
Would like to get the test_macos build time down but can't find a way to do so.  Perhaps it's only wise to run tests on macos before building the Pyinstaller distribution for Darwin?

hokusai/version.py
```diff
@@ -1,3 +1,4 @@
 import os
 
-VERSION = open(os.path.join(os.path.dirname(__file__), "VERSION")).read().strip()
```
izakp, line 3: Python 3 warns you about un-closed file descriptors during testing :)

izakp:
> I don't know python well enough for a very-specific review, but conceptually this all makes sense to me.

@eessex cool, I had not done a Python 2 to 3 migration before myself so was learning about the idiosyncrasies between the two as I went.

Also I just stumbled across another issue with formatting exceptions as I run this hokusai on this branch (via `pip install -e .` against Python 3.5.8) locally, so likely there are going to be further hidden issues.  I think it's more important to get this thoroughly beta-tested before we start shipping the Pyinstaller / brew built against Python 3 as there's no problem with continuing with Python 2, so I think i'll create another `beta3` build we can use.

Makefile
```diff
@@ -25,9 +25,15 @@ integration:
 
 test-docker:
 	$(DOCKER_RUN) \
-	  --volume "$(PWD)":"/src/$(PROJECT):ro" \
+	  --volume "$(PWD)":"/src/$(PROJECT):rw" \
```
dblandin, line 28: I think the default is `rw` so if we no longer want this to be a read-only volume mount, we could remove the suffix altogether.

izakp:
Now testing and building betas on both python 2 and 3 - canonical build is still built with python 2.

izakp:
Going to merge this and build the betas!

Issue #206, opened 2020-06-04
PyInstaller build not linked against correct openssl libs


izakp:
The PyInstaller build sometimes fails to make https requests with `SSL: CERTIFICATE_VERIFY_FAILED` this can happen on MacOS when it is not linked against brew-installed openssl libs.

We likely need to explicitly bundle the certificates with the python interpreter shipped by pyinstaller https://github.com/pyinstaller/pyinstaller/wiki/Recipe-OpenSSL-Certificate

Related issue? https://github.com/borgbase/vorta/issues/452

PR #214, opened 2020-07-06
fix ssl linking in pyinstaller builds

izakp:
Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-2543 and https://github.com/artsy/hokusai/issues/206 - fixes the Pyinstaller build so it includes the correct homebrew-installed openssl libs

Tested on the macos Pytinstaller build.

@artsyjian once merged can you test the beta build on Linux?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,7 +100,7 @@ jobs:
       xcode: "9.0"
     steps:
       - checkout
-      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install awscli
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -100,7 +100,7 @@ jobs:
       xcode: "9.0"
     steps:
       - checkout
-      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install awscli
+      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install awscli openssl
```
izakp, line 103: Wonder if we need to `apt-get install libssl-dev` for the Linux builds as well...

PR #217, opened 2020-07-13
replace pipenv with poetry

izakp:
Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-2542

Replaces Pipenv with Poetry for dev dependency management.  This decouples the management of virtualenvs from project dependency management, which are defined in `pyproject.toml` and locked in `poetry.lock`.  Poetry options are configured in `poetry.toml`.

Poetry needs to be installed via `pip`.  Installing project dependencies is as simple as `poetry install --no-root`.  Both these commands can be run with `make dependencies`.

Updating the lockfile is performed with `poetry lock`.

A developer can choose their preferred virtualenv workflow, or choose to ignore it all together.  In CI we don't need the extra overhead of working with virtualenvs and then running commands in them via `pipenv run`.

In my case I removed all Pipenv-created virtualenvs for the project with then set up a new virtualenv simply named `hokusai` with:

```
lsvirtualenv | grep hokusai- | xargs rmvirtualenv
mkvirtualenv hokusai
```

Then followed the documented steps to install poetry and dependencies:
```
workon hokusai
make dependencies
```

At this point you can run `ipython` and `import hokusai` successfully.

To install the projet in editable mode you can run `pip install -e .` and then in another directory run `workon hokusai` to use your local checkout of hokusai rather than any other installed distribution.

See https://github.com/artsy/ops-util/pull/5 for a related PR about migrating from pipenv to poetry.


izakp:
Adding `#HOLD` as this surfaced a build error - need to loosen the project python version dependency.

```
poetry install
The currently activated Python version 2.7.18 is not supported by the project (2.7.16).
Trying to find and use a compatible version. 

[NoCompatiblePythonVersionFound]
Poetry was unable to find a compatible version. If you have one, you can explicitly use it via the "env use" command.
make: *** [Makefile:15: dependencies] Error 1

Exited with code exit status 2
```

izakp:
Interesting note - poetry will install hokusai as it detects it is being run in a project with a `setup.py` file, unless `poetry install` is invoked with the `--no-root` option, but it will install the root project from github in editable mode, so added instructions to run `poetry install --no-root` then run `pip install -e .` if you want to install hokusai in editable mode from the local directory.

izakp:
Removed #HOLD - this PR should be good to merge.

PR #219, opened 2020-08-10
use the test or development yaml template when building in test or dev environments

izakp:
Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-2672

This ensures that build settings / target specified in test.yml or dev.yml are used when building the app image in test / dev mode, so for example an intermediary build target is produced in multistage builds rather than a production or default build target specified in build.yml

### DIFF
--- a/hokusai/commands/development.py
+++ b/hokusai/commands/development.py
@@ -28,7 +28,7 @@ def dev_start(build, detach, filename):
 
   opts = ''
   if build:
-    Docker().build()
+    Docker().build(filename=yaml_template)
   if detach:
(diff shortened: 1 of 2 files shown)


izakp:
There's no spec file for https://github.com/artsy/hokusai/blob/master/hokusai/services/docker.py as its shelling out to docker-compose but I think in the future it's worth expanding test suites to cover mocking the output of commands to docker / docker-compose and kubectl clis (this is probably worth its own spike though as we'd need some boilerplate / mocks in place globally when running the test suite).

PR #230, opened 2020-08-19
update build dependencies in Dockerfile

izakp:
Follow-up to https://github.com/artsy/hokusai/pull/229

Updates build dependencies for Docker


eessex:
thanks for this follow up!

PR #232, opened 2020-10-23
Fixes to exception and tempfile handling

izakp:
- Fixes the exception `AttributeError: 'NoneType' object has no attribute 'decode'` when catching an exception with no message or output attribute
- Fixes running hokusai in two terminal windows concurrently stomping on each other's temp directories
- Updates README.md with a 1-liner to install Hokusai via curl / bash

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,11 @@ $ pip uninstall hokusai
 $ brew link --overwrite hokusai
 ```
 
+If you are on Linux:
+```
+curl -sSL https://raw.githubusercontent.com/artsy/hokusai/master/get-hokusai.sh | sudo bash
+```
+
 ### Alternate Installation Methods
 
 #### Via pip
--- a/hokusai/services/yaml_spec.py
+++ b/hokusai/services/yaml_spec.py
@@ -1,5 +1,6 @@
 import os
 import atexit
+from tempfile import NamedTemporaryFile
 
 import yaml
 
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -128,6 +128,11 @@ $ pip uninstall hokusai
 $ brew link --overwrite hokusai
 ```
 
+If you are on Linux:
+```
+curl -sSL https://raw.githubusercontent.com/artsy/hokusai/master/get-hokusai.sh | sudo bash
```
dblandin, line 133: **nit (non-blocking)**: FYI, there's also a `docs` commit type (vs `chore`) commonly used documentation updates like this.

Didn't know we had this handy install script! 🎉

hokusai/services/yaml_spec.py
```diff
@@ -44,7 +45,7 @@ def to_file(self):
     file_basename = os.path.basename(self.template_file)
     if file_basename.endswith('.j2'):
       file_basename = file_basename.rstrip('.j2')
-    f = open(os.path.join(HOKUSAI_TMP_DIR, file_basename), 'w')
+    f = NamedTemporaryFile(delete=False, dir=HOKUSAI_TMP_DIR, mode='w')
```
dblandin, line 48: 👍

README.md
izakp: ooo will use `docs` next time :)

Issue #228, opened 2020-08-18
hoksuai check fails if project uses Jinja2 templates


izakp:
The `hokusai check` command looks for only `.yml` files but should succeed if the project uses `.yml.j2` / `.yaml.j2` templates as per [TemplateSelector](https://github.com/artsy/hokusai/blob/master/hokusai/lib/template_selector.py) logic

PR #235, opened 2020-10-27
fix: check command uses TemplateSelector

izakp:
Addresses https://github.com/artsy/hokusai/issues/228

### DIFF
--- a/hokusai/commands/check.py
+++ b/hokusai/commands/check.py
@@ -5,11 +5,12 @@ import botocore.exceptions as botoexceptions
 
 from hokusai import CWD
 from hokusai.lib.command import command
-from hokusai.lib.config import HOKUSAI_CONFIG_DIR, config
+from hokusai.lib.config import HOKUSAI_CONFIG_DIR, BUILD_YAML_FILE, TEST_YML_FILE, DEVELOPMENT_YML_FILE, config
 from hokusai.services.ecr import ECR
 from hokusai.services.kubectl import Kubectl
 from hokusai.lib.common import get_region_name, print_red, print_green, shout
 from hokusai.lib.exceptions import CalledProcessError, HokusaiError
+from hokusai.lib.template_selector import TemplateSelector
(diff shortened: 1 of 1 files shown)


alexpopa-artsy:
LGTM, tested `hokusai check` with Induction (uses .j2) and Gravity (.yml)

hokusai/commands/setup.py
```diff
@@ -63,14 +63,14 @@ def setup(project_name, template_remote, template_dir, template_vars, allow_miss
...
-  required_templates = [
+  builtin_templates = [
     'Dockerfile.j2',
     '.dockerignore.j2',
-    'hokusai/build.yml.j2',
-    'hokusai/development.yml.j2',
-    'hokusai/test.yml.j2',
-    'hokusai/staging.yml.j2',
-    'hokusai/production.yml.j2'
+    'hokusai/build.yml.j2.j2',
```
eessex, line 69: The double extension seems a little counter-intuitive to me, mostly because it is not a valid one as far as I can tell. I wonder if it makes more sense to add the suffix when parsing instead of stripping?

izakp: Ach @eessex I didn't mean to include those commits in this PR but as long as we are on the topic... yeah I also don't like these `.j2.j2` extensions but it does make sense to me in that these are j2 templates that render to j2 templates, and a render of the template should strip the trailing `.j2` file extension.

In essence, we should eventually do the same in our artsy-hokusai-templates repo and I was trying to preserve forwards compatibility with what we currently have there in so that `.j2` templates become regular files and `.j2.j2` templates become `.j2` files when running `hokusai setup`.  I am not sure how we could add the suffix when parsing and preserve all current compatibility while being able to migrate forward to this pattern and end up with `.yml.j2` files in the repo.

In any case I will remove these commits and open a separate PR where we can discuss this.

eessex (approved):

PR #236, opened 2020-11-03
#WIP #FF fea: Log config changes.

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-2958
Supported by: https://github.com/artsy/infrastructure/pull/257

To demonstrate the idea of logging config changes.

Every run of `hokusai...env set/unset...` generates a log like these (values may be sensitive so are omitted):
```
time: 2020-11-03 00:15:32.537098, namespace: default, action: set, var: FOO
time: 2020-11-03 00:16:10.925644, namespace: default, action: unset, var: FOO
```
and ships it to S3:
Screenshot from 2020-11-02 19-57-37

Might these logs be useful?

The implementation is not ideal. S3 files are clunky to work with, and updating them adds ~1 second delay to Hokusai's response, which is noticeable. If we consider config changes also as `releases`, perhaps better to let Horizon manage storage (in a DB), so that Hokusai can just POST these events to Horizon? And later on we make them retrievable via a `hokusai...env log` command?

If we do go with this implementation, I shall add tests.

### DIFF
--- a/hokusai/commands/env.py
+++ b/hokusai/commands/env.py
@@ -31,6 +31,7 @@ def set_env(context, environment, namespace=None):
     split = s.split('=', 1)
     configmap.update(split[0], split[1])
   configmap.save()
+  # TODO: pass to log_configmap_changes only var names, not their values.
   log_configmap_changes(context, 'set', environment)
 
 @command()
--- a/hokusai/commands/env.py
+++ b/hokusai/commands/env.py
@@ -32,7 +32,7 @@ def set_env(context, environment, namespace=None):
     configmap.update(split[0], split[1])
   configmap.save()
   # TODO: pass to log_configmap_changes only var names, not their values.
-  log_configmap_changes(context, 'set', environment)
+  log_configmap_changes(context, namespace, 'set', environment)
 
 @command()
 def unset_env(context, environment, namespace=None):
@@ -41,5 +41,5 @@ def unset_env(context, environment, namespace=None):
   for s in environment:
     configmap.delete(s)
   configmap.save()
-  log_configmap_changes(context, 'unset', environment)
+  log_configmap_changes(context, namespace, 'unset', environment)
 
--- a/hokusai/commands/env.py
+++ b/hokusai/commands/env.py
@@ -3,6 +3,7 @@ from hokusai.lib.command import command
 from hokusai.services.configmap import ConfigMap
 from hokusai.lib.common import print_green, print_smart
 from hokusai.lib.exceptions import HokusaiError
+from hokusai.lib.log_configmap_changes import log_configmap_changes
 
 
 @command()
@@ -30,7 +31,7 @@ def set_env(context, environment, namespace=None):
     split = s.split('=', 1)
     configmap.update(split[0], split[1])
   configmap.save()
-
+  log_configmap_changes(context, 'set', environment)
 
 @command()
 def unset_env(context, environment, namespace=None):
@@ -39,3 +40,5 @@ def unset_env(context, environment, namespace=None):
   for s in environment:
     configmap.delete(s)
   configmap.save()
+  log_configmap_changes(context, 'unset', environment)
+
--- a/hokusai/lib/log_configmap_changes.py
+++ b/hokusai/lib/log_configmap_changes.py
@@ -1,0 +1,30 @@
+import datetime
+
+from hokusai.lib.config import config
+from hokusai.lib.s3 import append_or_create_object
+
+def log_configmap_changes(context, action, varlist):
+  var_names = []
+  for s in varlist:
+    # if the vars are being set, varlist is a list of key=value pairs.
+    # if the vars are being unset, varlist is a list of keys.
+    # in either case, split returns the key.
+    split = s.split('=', 1)
+    var_names.append(split[0])
+
+  timestamp = datetime.datetime.utcnow()
+  log = "%s: %s: %s\n" % (timestamp, action, ' '.join(var_names))
+
+  # TODO: pass bucket/key in as a var.
+  s3_bucket = 'artsy-hokusai'
+  s3_key = 'configmap-changes/' + \
+            config.project_name + \
+            '/' + context + '/' + \
+            datetime.datetime.utcnow().strftime('%Y-%m-%d')
+
+  try:
+    # log to s3 object.
+    append_or_create_object(log, s3_bucket, s3_key)
+  except:
+    # catch and ignore all exceptions, failure to log configmap changes shall not be fatal.
+    pass
--- a/hokusai/lib/log_configmap_changes.py
+++ b/hokusai/lib/log_configmap_changes.py
@@ -3,17 +3,24 @@ import datetime
 from hokusai.lib.config import config
 from hokusai.lib.s3 import append_or_create_object
 
-def log_configmap_changes(context, action, varlist):
+def log_configmap_changes(context, namespace, action, varlist):
   var_names = []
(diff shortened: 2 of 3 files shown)


hokusai/commands/env.py
```diff
@@ -3,6 +3,7 @@
 from hokusai.services.configmap import ConfigMap
 from hokusai.lib.common import print_green, print_smart
 from hokusai.lib.exceptions import HokusaiError
+from hokusai.lib.log_configmap_changes import log_configmap_changes
```
artsyjian, line 6: Or should this be a `service` under `hokusai/services`?

hokusai/lib/log_configmap_changes.py
```diff
@@ -0,0 +1,37 @@
...
+  if (namespace is None):
+    ns = 'default'
+  else:
+    ns = namespace
+
+  timestamp = datetime.datetime.utcnow()
+  log = "time: %s, namespace: %s, action: %s, var: %s\n" % (timestamp, ns, action, ' '.join(var_names))
+
+  # TODO: pass bucket/key in as a var.
+  s3_bucket = 'artsy-hokusai'
```
artsyjian, line 26: not sure about the proper way to pass it in.

hokusai/commands/env.py
izakp: Services is somewhat motivated by integrations with other apis / programs such as Docker, Kubectl, etc.... though YamlSpec may be an exception it is a reference to a templating engine - yeah I think this may be a good case for putting it under services.

hokusai/lib/log_configmap_changes.py
izakp: I would hoist this up to config, so it can be set via a flag for a command or via an env var... see https://github.com/artsy/hokusai/blob/master/hokusai/lib/config.py

artsyjian: cool. thanks!

hokusai/commands/env.py
artsyjian: okay.

joeyAghion (commented):
Sorry for delay! This was compounded by the fact that I don't see perfect solutions to some of these challenges.

Generally, yes I think being able to trace through config changes is useful. I'm not in love with S3 as the destination because that's somewhat arbitrary and a significant new scope/dependency-increase for Hokusai. Logging the changes with our regular logging solution isn't desirable due to the sensitivity. Putting them in Horizon's database also seems like an arbitrary scope-increase for that system. I keep coming back to whether we can just use kubernetes's native concepts for this, like storing the config-map archives as config-maps (or maybe even within some meta-config-map). Or is there another handy key/value store?

izakp (commented):
@artsyjian as we discussed on Slack I think, would advise that we try and use a Kubernetes ConfigMap to log changes rather than an S3 bucket

dblandin:
> I keep coming back to whether we can just use kubernetes's native concepts for this, like storing the config-map archives as config-maps (or maybe even within some meta-config-map). Or is there another handy key/value store?

One thought is to use AWS Parameter Store. I've used it the past along with [segmentio/chamber](https://github.com/segmentio/chamber) to manage configuration for services running in Elastic Container Service.

**Some benefits**

- Fast
- CLI support
- AWS SDK support
- Versioning
- Can encrypt input

Would be great to stay as close to Kubernetes as possible. If there's a way to react whenever a ConfigMap is created/updated (via hokusai / UI dashboard / etc) and automatically persist the value into Parameter Store, that might be pretty cool. We could then use the Parameter Store APIs within hokusai to list config map versions and rollback to previous states.

**Resources**

- https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html
- https://docs.aws.amazon.com/cli/latest/reference/ssm/index.html#cli-aws-ssm

### Examples

**Setting a secure parameter with a JSON payload**

```
$ aws ssm put-parameter --name 'hokusai.force.staging' --type SecureString --value='{ "key1: "0" }'
```

**Retrieving a secure parameter with decryption**

```
$ aws ssm get-parameter --name 'hokusai.force.staging' --with-decryption
{
    "Parameter": {
        "Name": "hokusai.force.staging",
        "Type": "SecureString",
        "Value": "AQICAHhAgvzOnt3NYiFUmNvonq2rTa69MkqjugF47bK/kLyYYQFUCejUHRDawH4uUlZAfCQtAAAAbTBrBgkqhkiG9w0BBwagXjBcAgEAMFcGCSqGSIb3DQEHATAeBglghkgBZQMEAS4wEQQMa5gT13X4nHXK0JwXAgEQgCoTZkT/X91vBrPSj13rdSfZwOQBfLAanI5oUlBgiB0k87N2+2NClXbSnfc=",
        "Version": 4,
        "LastModifiedDate": "2021-02-19T11:03:37.820000-05:00",
        "ARN": "arn:aws:ssm:us-east-1:585031190124:parameter/hokusai.force.staging",
        "DataType": "text"
    }
}
```

**Parsing JSON from decrypted parameter**


```
$ aws ssm get-parameters --name 'hokusai.force.staging' --query "Parameters[0].Value" --with-decryption | jq '.|fromjson'
{
  "key1": "0"
}
```

**Retrieving parameter history**


```
$ aws ssm get-parameter-history --name 'hokusai.force.staging' --with-decryption | jq '.Parameters[] | { version: .Version, value: .Value }'
{
  "version": 1,
  "value": "{ \"key1\": \"0\" }"
}
{
  "version": 2,
  "value": "{ \"key1\": \"1\" }"
}
{
  "version": 3,
  "value": "{ \"key1\": \"3\" }"
}
{
  "version": 4,
  "value": "{ \"key1\": \"3\", key2: \"0\" }"
}
```

artsyjian:
Thanks @devon for the idea. The concept is really cool. If we want configs archived, a non-k8s key/value store (AWS Parameter Store, HashiCorp Vault, ...) makes a lot of sense.

@isac, @joey - About using k8s configmap concept, it seems to depend on whether we want changelogs or archives.

For logging, configmap should work. I imagine each `app-environment` configmap can have a buddy `app-environment-changes` configmap. When `FOO=bar` is added (or deleted/modified) to `app-environment`, we can append to `app-environment-changes` with a line like this:
```
time: 2020-11-03 00:20:27.384867, user: banksy, action: set, var: FOO, value: bar
```

A configmap's size is [limited](https://kubernetes.io/docs/concepts/configuration/configmap/) at 1MB. If a line averages 200 bytes, `app-environment-changes` can store 5k deltas which should be years worth.

However, if we want config archives, we probably shouldn't use configmaps. Say we archive each old version of `app-environment` in a dedicated configmap, we will end up with hundreds or even 1k+ configmaps. At that number, we might encounter problems as someone [did](https://github.com/kubernetes/kubernetes/issues/74412). Or, if we pool all the old versions into one giant configmap, it won't hold enough revisions, due to the 1MB limit (force's configmap is ~20KB for example). The total storage size of old config versions across the entire cluster might also be a concern as [etcd](https://etcd.io/docs/v3.4.0/dev-guide/limit/) seems designed to not store too much data:

> Request size limit
> 
> etcd is designed to handle small key value pairs typical for metadata. Larger requests will work, but may increase the latency of other requests. By default, the maximum size of any request is 1.5 MiB. This limit is configurable through --max-request-bytes flag for etcd server.
> Storage size limit
> 
> The default storage size limit is 2GB, configurable with --quota-backend-bytes flag. 8GB is a suggested maximum size for normal environments and etcd warns at startup if the configured value exceeds it.
>

Issue #237, opened 2020-11-09
Docker login error prints credentials in the console


starsirius:
Perhaps, as suggested in the screenshot, we can switch to use `--password-stdin`.

---

![Screen Shot 2020-11-09 at 3 34 41 PM](https://user-images.githubusercontent.com/796573/98598046-e2823900-22a7-11eb-9ffe-067d212cda03.png)

PR #238, opened 2020-11-10
fea: mask printing errors when shelling out to a subcommand

izakp:
Addresses https://github.com/artsy/hokusai/issues/237

We already mask verbose output if the kwarg `mask` is provided to a subcommand - this fix wraps `CalledProcessError`s in our custom exception class `HokusaiError` and applies the mask to `HokusaiError` `message` attribute if provided.

### DIFF
--- a/hokusai/lib/common.py
+++ b/hokusai/lib/common.py
@@ -17,7 +17,7 @@ from botocore import session as botosession
 from termcolor import cprint
 
 from hokusai.lib.config import config
-from hokusai.lib.exceptions import CalledProcessError
+from hokusai.lib.exceptions import HokusaiError, CalledProcessError
 
 CONTEXT_SETTINGS = {
   'terminal_width': 10000,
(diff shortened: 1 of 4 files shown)


artsyjian (approved):
LGTM

izakp:
Hmm seeing some failures on Python 3 tests...

izakp:
Okay was a transient dependency error - I think this is fine to merge.

hokusai/lib/common.py
```diff
@@ -78,10 +78,13 @@ def returncode(command, mask=()):
...
-    return check_call(verbose(command, mask=mask), stderr=STDOUT, shell=True)
-  else:
-    return check_output(verbose(command, mask=mask), stderr=STDOUT, shell=True)
+  try:
+    if print_output:
+      return check_call(verbose(command, mask=mask), stderr=STDOUT, shell=True)
+    else:
+      return check_output(verbose(command, mask=mask), stderr=STDOUT, shell=True)
+  except CalledProcessError as e:
+    raise HokusaiError(str(e), return_code=e.returncode, mask=mask)
```
izakp, line 87: Passing the string representation of the [CalledProcessError](https://docs.python.org/2/library/subprocess.html#subprocess.CalledProcessError) rather than its output attribute as this may be `None` - also we want to ensure a string rather than unicode when it is caught and its message attribute is printed in `command.py`

izakp: Actually seeing some perhaps unexpected changes to hokusai's behavior due to these changes... This introduces a change that the `CalledProcessError`'s returncode is passed through to `HokusaiError` - this effectively changes hokusai's behavior in that it will return the returncode of the error's subcommand rather than `1` - I was surprised we don't do this already - we are currently catching all CalledPRocessErrors separately [here](https://github.com/artsy/hokusai/blob/master/hokusai/lib/command.py#L29) and returning `1`.  Perhaps this exception handling needs to be refactored...

izakp:
Actually not sure this is ok to merge... maybe the exception handing in `command.py` needs to be refactored... see my comment above ^

izakp:
Actually seeing that this may introduce odd behavior - we catch `CalledProcessErrors` as part of control flow in other places and suppressing these exceptions naturally bubbling up will break this control flow like [here](https://github.com/artsy/hokusai/blob/master/hokusai/services/deployment.py#L184)

PR #239, opened 2020-11-10
fix: mask CalledProcessError msg and output attributes and treat HokusaiError separately from CalledProccessError on exit

izakp:
Another approach following https://github.com/artsy/hokusai/pull/238

Addresses https://github.com/artsy/hokusai/issues/237

We already mask verbose output if the kwarg `mask` is provided to a subcommand via the `shout` function - this fix also applies the mask to a `CalledProcessessError`'s `cmd` and `output` attributes, should one be raised.

Also fix exception handling in the `@command` decorator and treat `HokusaiError`s separately from `CalledProcessError`s - branching for Python 2/3 compatibility.

### DIFF
--- a/hokusai/lib/command.py
+++ b/hokusai/lib/command.py
@@ -19,22 +19,34 @@ def command(config_check=True):
           sys.exit(0)
         else:
           sys.exit(result)
-      except HokusaiError as e:
-        print_red(e.message)
-        sys.exit(e.return_code)
       except SystemExit:
         raise
       except KeyboardInterrupt:
         raise
-      except (CalledProcessError, Exception) as e:
+      except HokusaiError as e:
+        print_red("ERROR: %s" % e.message)
+        sys.exit(e.return_code)
+      except CalledProcessError as e:
         if get_verbosity() or os.environ.get('DEBUG'):
           print_red(traceback.format_exc())
         else:
           print_red("ERROR: %s" % str(e))
         if hasattr(e, 'output') and e.output is not None:
-          print(e.output.decode('utf-8'))
-        elif hasattr(e, 'message') and e.message is not None:
-          print(e.message.decode('utf-8'))
+          if str == bytes:
+            # Python 2
+            print_red("ERROR: %s" % e.output)
+          else:
+            # Python 3
+            if type(e.output) == bytes:
+              print_red("ERROR: %s" % e.output.decode('utf-8'))
+            else:
+              print_red("ERROR: %s" % e.output)
+        sys.exit(e.returncode)
+      except Exception as e:
+        if get_verbosity() or os.environ.get('DEBUG'):
+          print_red(traceback.format_exc())
+        else:
+          print_red("ERROR: %s" % str(e))
         sys.exit(1)
     return wrapper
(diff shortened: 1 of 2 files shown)


izakp:
I tested this locally by causing the call to `docker login` to fail...
```
(hokusai) isacpetruzzi@Artsy-Isac hokusai-sandbox (master) $ DEBUG=1 hokusai registry push --force
==> hokusai exec `dockerz login -u ****** -p ***** https://585031190124.dkr.ecr.us-east-1.amazonaws.com`

  File "/Users/isacpetruzzi/src/artsy/hokusai/hokusai/lib/command.py", line 17, in wrapper
    result = func(*args, **kwargs)
  File "/Users/isacpetruzzi/src/artsy/hokusai/hokusai/commands/push.py", line 22, in push
    shout(ecr.get_login(), mask=(r'^(dockerz login -u) .+ (-p) .+ (.+)$', r'\1 ****** \2 ***** \3'))
  File "/Users/isacpetruzzi/src/artsy/hokusai/hokusai/lib/common.py", line 85, in shout
    return check_output(verbose(command, mask=mask), stderr=STDOUT, shell=True)
  File "/Users/isacpetruzzi/.pyenv/versions/2.7.16/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 223, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
CalledProcessError: Command 'dockerz login -u ****** -p ***** https://585031190124.dkr.ecr.us-east-1.amazonaws.com' returned non-zero exit status 127

ERROR: /bin/sh: dockerz: command not found
```

Will add a test around this as a follow-up.

jpotts244 (approved):
A+ catch on the error output here! 🚢

hokusai/lib/command.py
```diff
@@ -19,22 +19,34 @@ def wrapper(*args, **kwargs):
...
+      except CalledProcessError as e:
         if get_verbosity() or os.environ.get('DEBUG'):
           print_red(traceback.format_exc())
         else:
           print_red("ERROR: %s" % str(e))
         if hasattr(e, 'output') and e.output is not None:
-          print(e.output.decode('utf-8'))
-        elif hasattr(e, 'message') and e.message is not None:
-          print(e.message.decode('utf-8'))
+          if str == bytes:
```
jpotts244, line 35: I am sorry there has to be conditionals based on python versions! 😭 Nice catch on handling both.

Not blocking but maybe someday we can encapsulate some of the python checks out into their own function since it is needed in a few places.

izakp: Yeah the string / bytes unicode changes between Python 2 / 3 are doing my head in 😓   This seems to be the most concise way to check it.  Good call on a single function... i.e. `which_python()` that returns `2` or `3`.  That would make my life a lot easier!

dblandin (approved):

starsirius (approved):
Thank you for the quick fix @izakp ! Feel free to merge when ready (and +1 to add a test for this as a followup).

izakp:
Ok merging - we can try a beta build in the meantime

PR #240, opened 2020-11-16
Install JQ in hokusai image

starsirius:
JQ is required for the new version of [Slack Orb](https://github.com/CircleCI-Public/slack-orb/wiki/FAQ). The size is [< 1MB](https://pkgs.alpinelinux.org/package/edge/main/x86/jq) so hopefully not a lot of overhead. 🙏


izakp:
LGTM!

PR #244, opened 2020-12-14
feature: Hokusai Lite

icirellik:
Removes the docker daemon from the `artsy/hokusai` image reducing it's
size by over 30%. This will decrease the executor spin up time in
CircleCI for all repo using hokusai.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,20 +3,20 @@ FROM python:2.7-alpine
 WORKDIR /src
 
 # Install Docker
-RUN apk add \
-      build-base \
-      libffi-dev \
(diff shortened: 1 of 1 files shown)


icirellik:
@izakp  I'm been meaning to look into this image for a while now. Let me know if we there is anywhere we use the docker daemon locally and I will create a separate `lite` image for specific use cases. Everywhere I look checked appeared to connect to a remote docker daemon.

izakp:
Excellent @icirellik !  Yeah we use a remote docker daemon everywhere this container is run.

izakp (approved):

Issue #248, opened 2021-02-08
Hokusai setup in the docs


fclesio:
I was trying to configure Hokusai using the [docs](https://github.com/artsy/hokusai#setup) and I found out that the docs aren't complete. 

For example: In the docs says:
- 1) AWS Credentials,  
- 2) Run the `hokusai configure --kubectl-version <kubectl version> --s3-bucket <bucket name> --s3-key <file key>` command. 

However, running that command of the step 2, I got the following error:

![Screenshot from 2021-02-08 15-32-02](https://user-images.githubusercontent.com/10605378/107233922-640e7200-6a23-11eb-9c3d-835d5b04d470.png)

I think the `hokusai setup` should come before than the configure, no?

dblandin:
@fclesio Thanks for the issue! I think that's actually a bug and not a documentation issue.

- `hokusai configure` is meant to configure hokusai dependencies globally like installing kubectl and pulling down kubernettes configuration files from S3.
- `hokusai setup` is for setting up a project to use hokusai

`hokusai setup` currently skips this check (because the project configuration files do not exist yet) but `hokusai configure` does not and I think that might be an oversight because I don't see anything in the command definition that relies on that project-specific context. Seems like you should be able to run `hokusai configure` anywhere on your system.

I've opened https://github.com/artsy/hokusai/pull/249 to enable the same `config_check=False` flag for the `configure` command.

izakp:
Closed via #249

PR #249, opened 2021-02-19
fix: skip project config check when configuring hokusai

dblandin:
The `hokusai configure` command globally configures hokusai by installing kubectl and fetching kubernetes configuration files from S3.

The command does not need to run within the context of a hokusai-configured project so I think we can safely skip this check like we currently do for `setup` and `version` commands.

Addresses #248


artsyjian (approved):
LGTM. Thanks @dblandin!

izakp (approved):
Great!  This has been a bit of a nuisance - perhaps (minor comment) an update in the docs  to  reflect this - otherwise LGTM!

PR #254, opened 2021-05-24
Izakp/further python3 fixes

izakp:
Found more errors on Python 3

- Need to decode the binary output of subprocess return values on Python 3 -  fix and add unit test

Will keep this #WIP as I hunt for further Python 3 bugs


artsyjian (approved):
:+1:

PR #256, opened 2021-06-17
fix: reverted promt-toolkit version due to mismatch with python 3.5

aryartsy:
* reverted prompt-toolkit version to 1.0.16 since later versions are not compatible with python 3.5 [https://pypi.org/project/prompt-toolkit/#history](url)
* this change may remove the possibility to install ipdb / ipython in dev mode which was added here: [https://github.com/artsy/hokusai/pull/254](url)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(name='hokusai',
       install_requires=[
           'click~=6.7',
           'click-repl~=0.1',
-          'prompt-toolkit~=3.0.18',
+          'prompt-toolkit~=1.0.16',
           'MarkupSafe~=1.0',
           'Jinja2~=2.10',
           'packaging',


alexpopa-artsy (approved):
LGTM!

alexpopa-artsy:
@artsyjian for a bit more context, this PR should fix the issue arising [here](https://app.circleci.com/pipelines/github/artsy/hokusai/446/workflows/32c4577c-d9bf-430c-9486-d74985e37522/jobs/1724). The issue was that that specific version of prompt-toolkit does not have python 3.5.x support, so the tests were failing.
@aryartsy correct me if I'm wrong, but we were unable to trigger the `release_*` tests to check completely. I assume those test only run after a PR is merged.

Once again, thanks @aryartsy for taking the initiative to fix this!

artsyjian:
@aryartsy, @alexpopa-artsy - Thanks for looking into this. Do you know if reverting`prompt-toolkit` affects anything else in this [PR](https://github.com/artsy/hokusai/pull/254)? The [commit](https://github.com/artsy/hokusai/pull/254/commits/889cac4fd5091c25db303606cef626dbc4e64165) says it's for installing `ipdb` and `ipython` in dev mode. Do you know if we don't need that anymore? (I don't recall what they were required for).

alexpopa-artsy:
@artsyjian I have no idea, but I guess if the tests pass it should be ok?
maybe @aryartsy has a clue

aryartsy:
@artsyjian & @alexpopa-artsy according to their documentation prompt_toolkit is a library for building interactive command line applications in Python.
I didn't find any related changes or usage in this repo related to prompt-toolkit library, and we had no release with that version change either.

ovasdi:
Merging this to fix the `release_beta_dockerhub` job as we need to release a fix.

PR #257, opened 2021-06-21
doc: README added python version flag to mkvirtualenv

aryartsy:
- README instruction update: added missing python version flag when creating virtual environment. Not having this flag could cause to pick the system Python over the global Pyenv Python version, which eventually causes failure if the system version is not matching the accepted python version


alexpopa-artsy (approved):
LGTM!

PR #259, opened 2021-07-29
Fix make image by setting prompt-toolkit's version to bigger than 2.0.0

chr-tatu:
## Description
Currently, the make image CircleCI is failing because `prompt-toolkit` version 3.0.18 is not found. The error states that only versions lower than 3.0 are available.
When testing locally, `make image` passed successfully.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(name='hokusai',
       install_requires=[
           'click~=6.7',
           'click-repl~=0.1',
-          'prompt-toolkit~=3.0.18',
+          'prompt-toolkit >= 2.0.0',
           'MarkupSafe~=1.0',
           'Jinja2~=2.10',
           'packaging',


ovasdi:
I've merged the [PR](https://github.com/artsy/hokusai/pull/256) that was already open to revert [this change](https://github.com/artsy/hokusai/pull/254/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R31). This may not be necessary.

There is also [a ticket](https://artsyproduct.atlassian.net/browse/PLATFORM-3465) to move us off of python 2.7 dockerized builds.

chr-tatu:
Awesome!

> I've merged the [PR](https://github.com/artsy/hokusai/pull/256) that was already open to revert [this change](https://github.com/artsy/hokusai/pull/254/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R31). This may not be necessary.
> 
> There is also [a ticket](https://artsyproduct.atlassian.net/browse/PLATFORM-3465) to move us off of python 2.7 dockerized builds.

Ah, great! I didn't check the existing PR. Thanks, @ovasdi. I'll close this one then.

PR #269, opened 2021-09-30
update: default to main instead of master

kajatiger:
In order to go with the current standard we need hokusai to default to the main branch with its commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,9 @@ Hokusai is currently tested on Pythons 2.7.16 and 3.5.8.
 
 2) Install dependencies: `poetry install --no-root`.
 
-3) Run tests
+3) make sure you installed minikube with `brew install minikube` and run with `minikube start `
+
+4) Run tests
 
 All tests can be run with `make tests`.
 
--- a/hokusai/cli/production.py
+++ b/hokusai/cli/production.py
@@ -36,7 +36,7 @@ def delete(filename, verbose):
 
 
 @production.command(context_settings=CONTEXT_SETTINGS)
-@click.option('--check-branch', type=click.STRING, default=detect_branch(), help='Check branch before updating (default: main)')
+@click.option('--check-branch', type=click.STRING, default=detect_branch(), help='Check branch before updating (default: origin default or main)')
 @click.option('--check-remote', type=click.STRING, help='Check remotes before updating (otherwise check all remotes)')
 @click.option('--skip-checks', type=click.BOOL, is_flag=True, help='Skip all checks and update configuration recklessly')
 @click.option('-f', '--filename', type=click.STRING, help='Use the given Kubernetes Yaml file (default ./hokusai/production.yml)')
--- a/hokusai/cli/production.py
+++ b/hokusai/cli/production.py
@@ -34,7 +34,7 @@ def delete(filename, verbose):
 
 
 @production.command(context_settings=CONTEXT_SETTINGS)
-@click.option('--check-branch', type=click.STRING, default="master", help='Check branch before updating (default: master)')
+@click.option('--check-branch', type=click.STRING, default="main", help='Check branch before updating (default: main)')
 @click.option('--check-remote', type=click.STRING, help='Check remotes before updating (otherwise check all remotes)')
 @click.option('--skip-checks', type=click.BOOL, is_flag=True, help='Skip all checks and update configuration recklessly')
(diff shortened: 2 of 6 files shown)


hokusai/lib/default_branch.py
```diff
@@ -0,0 +1,9 @@
+from hokusai.lib.common import shout
+
+def detect_branch():
+  try: 
+    default_branch = shout("git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@'")
```
kajatiger, line 5: This git command will look up the local default HEAD, not the remote one. It seems like hokusai is always fetching remote before running automations and should already have the current version of the default branch also being mirrored locally. There is however another command with `git remote show` that could even make another call to the actual git remote in order to check live what is the current status of HEAD there. Under the circumstances it seems redundant though... Thanks @sonecabr for explaining this.

README.md
```diff
@@ -209,7 +209,9 @@ Hokusai is currently tested on Pythons 2.7.16 and 3.5.8.
 
 2) Install dependencies: `poetry install --no-root`.
 
-3) Run tests
+3) make sure you installed minikube with `brew install minikube` and run with `minikube start `
```
alexpopa-artsy, line 212: Nitpick: could you explain why `minikube` is necessary? I don't have `minikube` set up and never noticed any issues (but I also haven't developed within hokusai!)

hokusai/cli/production.py
```diff
@@ -4,6 +4,7 @@
 
 from hokusai.cli.base import base
 from hokusai.lib.common import set_verbosity, CONTEXT_SETTINGS
+from hokusai.lib.default_branch import detect_branch
```
alexpopa-artsy, line 7: Nitpick: it might make more sense to call this file something like `branch_detection.py`; `default_branch` is a bit confusing because the default branch is just one of the possible return values.

hokusai/lib/default_branch.py
alexpopa-artsy: Looking through `shout` and without any documentation there, it's not clear to me what it does and why. @kajatiger another way to execute such commands would be 
```
import os
os.system("git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@'")
```
but I strongly suspect `shout()` is the desired way, so all good!

In a local python console in another repo I get the following output:
```
>>> import os
>>> os.system("git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@'")
master
0
>>>
```
Guess that `0` is not desired so I think you're doing the right thing here!

hokusai/lib/default_branch.py
```diff
@@ -0,0 +1,9 @@
+from hokusai.lib.common import shout
+
+def detect_branch():
```
alexpopa-artsy, line 3: Update: could you add just a bit of documentation here, for future reference? Something like "This function detects the default main/master branch from the repo and returns it as string" or something like that.

alexpopa-artsy (approved):
LGTM! I haven't tested it locally, but syntactically looks good!

README.md
kajatiger: Yes it fixed the broken setup for me for some reason. I honestly don't understand why, but it was the only thing that made it work.

hokusai/lib/default_branch.py
kajatiger: We are already using `shout` in combination with running a git command in another place in hokusai, so this is where we took this from. I also ran the plain git command in 2 different projects one with "main" and one with "master" to check if it actually works as expected and it should be fine.

alexpopa-artsy:
Btw @kajatiger if I remember correctly, congrats on your first PR in python 🎉!

sonecabr (approved):
LGTM

hokusai/cli/staging.py
```diff
@@ -34,7 +35,7 @@ def delete(filename, verbose):
 
 
 @staging.command(context_settings=CONTEXT_SETTINGS)
-@click.option('--check-branch', type=click.STRING, default="master", help='Check branch before updating (default: master)')
+@click.option('--check-branch', type=click.STRING, default=detect_branch(), help='Check branch before updating (default: main)')
```
joeyAghion, line 38: Minor: the `default: main` help text isn't quite accurate. Maybe it should say "defaults to the origin remote's default branch" now.

kajatiger: You're right, I updated it 👍

joeyAghion (approved):

PR #276, opened 2021-12-07
feat: add image digest support to CommandRunner:run()

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-3772

This change adds _image digest_ support as a function argument to the `CommandRunner:run()` function. The calling code can now pass either a `tag` or a `digest` and the `run()` function will build a valid _image_name_.

The change modifies the _pre-deploy_ and _post-deploy_ hooks within `Deployment:update()` function where an image `tag` is being used to build an _image_name_ instead providing a `digest` to ensure that the same image is used throughout the whole deploy process.

- when a _staging_ `tag` is passed, the _image_name_ is: `123.dkr.ecr.us-east-1.amazonaws.com/some-app:staging`
- when a _sha256:111_ `digest` is passed, the _image_name_ is: `123.dkr.ecr.us-east-1.amazonaws.com/some-app@sha256:111`

This change mitigates the likelihood that a wrong image will be used during the deploy process if a staging deploy is triggered and the `staging` tag is updated during an ongoing production deploy.

### DIFF
--- a/hokusai/services/command_runner.py
+++ b/hokusai/services/command_runner.py
@@ -15,7 +15,7 @@ class CommandRunner(object):
     self.kctl = Kubectl(self.context, namespace=namespace)
     self.ecr = ECR()
 
-  def run(self, image_tag, cmd, tty=None, env=(), constraint=()):
+  def run(self, tag_or_digest, cmd, tty=None, env=(), constraint=()):
     if os.environ.get('USER') is not None:
       # The regex used for the validation of name is '[a-z0-9]([-a-z0-9]*[a-z0-9])?'
       user = re.sub("[^0-9a-z]+", "-", os.environ.get('USER').lower())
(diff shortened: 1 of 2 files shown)


artsyjian:
Wow, very interesting problem! This fix looks good - it ensures that the same docker image is used throughout `Deployment.update()`.

`Deployment.update()` is called from [pipeline promote](https://github.com/artsy/hokusai/blob/1fe7d101425c1c529d0a17420c2ea6c90ec24944/hokusai/commands/deployment.py#L28) which at first may call `CommandRunner` to run a migration, supplying the tag that will be used in `Deployment.update()`. Is there a concern for race condition there? Or is migration only ever run manually?

artsyjian (approved):
lgtm

joeyAghion (approved):
@artsyjian yes a similar risk exists if the `--migration` argument is passed to the `pipeline promote` subcommand. We didn't address that just because we didn't think that argument was in use anywhere.

sonecabr (approved):
LGTM

PR #283, opened 2022-03-16
chore: switch test_integration image to a supported one

ovasdi:
CircleCI sent an email regarding deprecation of machine images.

<details><summary>Email Screenshot</summary>

Screen Shot 2022-03-16 at 4 18 24 PM

</details>

Hokusai currently uses the `circleci/classic:201808-01` in the test_integration job. I located [this blog post](https://circleci.com/blog/ubuntu-14-16-image-deprecation/) which states that in addition to 14.04 images, 16.04 images are also affected by this. Specifically the section "_Does the Ubuntu 14.04 and 16.04 image deprecation affect me?_" (includes the image tag we use):

> In addition, the following Ubuntu 16.04-based images will be removed:
>
> circleci/classic:201808-01 <-- THIS IMAGE
> ubuntu-1604:201903-01
> ubuntu-1604:202004-01
> ubuntu-1604:202007-01
> ...

Updated the config following the [suggested migration steps](https://circleci.com/docs/2.0/images/linux-vm/16.04-to-20.04-migration/#switching-to-a-supported-image) to use a supported image instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
       - run:
           name: install conntrack
           command: |
-            apt-get -qq update && apt-get -qq install conntrack
+            sudo apt-get -qq update && sudo apt-get -qq install conntrack
       - run:
           name: start minikube
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,11 @@ jobs:
           name: setup minikube
           command: |
             curl -Lo minikube https://github.com/kubernetes/minikube/releases/download/${MINIKUBE_VERSION}/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
+      # required by minikube for k8s v1.18 onwards
+      - run:
+          name: install conntrack
+          command: |
+            apt-get -qq update && apt-get -qq install conntrack
       - run:
           name: start minikube
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ jobs:
 
   test_integration:
     machine:
-      image: circleci/classic:201808-01
+      image: ubuntu-2004:202201-02
     environment:
       K8S_VERSION: v1.17.17
       KUBECONFIG: /home/circleci/.kube/config
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,6 +65,7 @@ jobs:
           name: Update pyenv
           command: |
             cd /opt/circleci/.pyenv/plugins/python-build/../.. 
+            git checkout master
             git pull
             cd -
       - run: pyenv install 2.7.16
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   test_linux2:
     docker:
-      - image: python:2.7.16
+      - image: python:2.7
     steps:
       - checkout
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -65,6 +65,7 @@ jobs:
           name: Update pyenv
           command: |
             cd /opt/circleci/.pyenv/plugins/python-build/../.. 
+            git checkout master
```
ovasdi, line 68: Seems with the new image, pyenv comes with the HEAD at v2.2.4:
```
circleci@default-12f749eb-c35b-4e89-8e46-dff291662509:/opt/circleci/.pyenv$ git status
HEAD detached at v2.2.4
```

Added a checkout to remedy this.

ovasdi: This update and install step adds 2-3 mins. The new image comes iwth 2.7.18. Wonder if we could just use that version and drop this install.

artsyjian: the version of pyenv? looks like latest is 2.2.5.
```
jian@artsy:~/internetcode/pyenv$ git tag | grep v2.2 | tail -1
v2.2.5
```
so perhaps 2.2.4 is fine.

.circleci/config.yml
```diff
@@ -19,9 +19,9 @@ jobs:
 
   test_integration:
     machine:
-      image: circleci/classic:201808-01
+      image: ubuntu-2004:202201-02
     environment:
-      K8S_VERSION: v1.17.17
+      K8S_VERSION: v1.18.10
```
artsyjian, line 24: didn't know a version is coded here. thought it'd be good to match prod/stg.

artsyjian (approved):
thanks @ovasdi for catching the circleci deprecation.

.circleci/config.yml
ovasdi: Hmm, tried rebuilding and failed again with:

```
#!/bin/bash -eo pipefail
sudo -E minikube start --vm-driver=none --cpus 2 --memory 2048 --kubernetes-version=${K8S_VERSION} &> $HOME/minikube.log 2>&1 < /dev/null


Exited with code exit status 86
CircleCI received exit code 86
```

Looking up the [error code](https://minikube.sigs.k8s.io/docs/contrib/errorcodes/#error-codes-specific-to-the-guest-host): `86: ExGuestUnsupported`
Unclear to me what the issue is. Perhaps minikube version needs a bump too?

artsyjian: interesting! this [doc](https://minikube.sigs.k8s.io/docs/contrib/errorcodes/) says about the error code:
```
“GUEST_MISSING_CONNTRACK” (Exit code ExGuestUnsupported)
minikube could not find conntrack on the host, which is required from Kubernetes 1.18 onwards
```

ovasdi: Nice.

artsyjian: [resolved](https://app.circleci.com/pipelines/github/artsy/hokusai/543/workflows/a61db686-357f-4b12-8c70-d2f6cd51737c)!

ovasdi:
The integration step in the last run is 2mins faster after _removing_ the pyenv upgrade and installation of python 2.7.16, as expected.
How do we feel about this getting merged and using python 2.7.18 going forward?

artsyjian:
> How do we feel about this getting merged and using python 2.7.18 going forward?

I feel fine. We should probably drop Python2 altogether.

PR #289, opened 2022-04-08
require all tests before beta, update get-hokusai script.

artsyjian:
Minor improvements:

- require all tests to pass before releasing any beta
- `brew update` is required to detect latest version of Homebrew formula
- `sudo` is required for `get-hokusai.sh` to install hokusai binary into `/usr/local/bin`
- instruct to use `get-hokusai.sh` to install beta


ovasdi (approved):
🦖

PR #290, opened 2022-04-08
fix beta download command

artsyjian:
- revert `get-hokusai.sh` script. the readme already says to use `sudo` so no need to embed it in script.
- fix beta download command.


ovasdi (approved):
🦖

Issue #291, opened 2022-04-12
track app version in an env var


artsyjian:
It seems reasonable that a Kubernetes deployment's pods have an Env var that tracks the version of the app that's running.

For example, a pod of `example-app` deployment can have a var called `APP_VERSION` that tracks the version of the app that the deployment was rolled out with. In this case, the version is represented by the digest of the Docker image being run.
```
$ kubectl exec -it example-app-d5fc58d65-szfn9 env | grep APP_VERSION
APP_VERSION=sha256:0e454524f5b66b806f091fbba6a388082eb3831dca34da2a873fc348bf9a9e52
```
Having such a var is useful for folks who use an Application Performance Monitoring (APM) service such as Datadog whose trace agent automatically looks for the app version ([in DD_VERSION var by default](https://docs.datadoghq.com/tracing/setup_overview/setup/ruby/#environment-and-tags)) in Env. The metrics emitted are labeled with the version. That association is [visible on their UI's graphs](https://docs.datadoghq.com/tracing/deployment_tracking/) - metrics from different versions are colored differently. That helps the user correlate app metric (request rate, error rate) changes with releases.

This issue asks that:
- The following Hokusai commands update an Env var with the app version being deployed.
  - `hokusai [staging|production] deploy`
  - `hokusai pipeline promote`
- The value stored in that var is the digest of the Docker image that was deployed.
- The name of the var is configurable in a project's `./hokusai/config.yml`.

PR #292, opened 2022-04-12
track app version

artsyjian:
Resolves: https://github.com/artsy/hokusai/issues/291


joeyAghion (commented):
I know we discussed this already, but I don't remember the outcome and I know you did a bunch of experimentation since then, so I'll repeat my thinking here.

It feels a little arbitrary for `hokusai` to inject a Datadog-specific environment variable into all the apps it orchestrates. The major reason _to do_ this would appear to be that all of our apps would get release-tracking with no further code/config changes. (Granted, this is a compelling one!)

The reasons _not to_ in my opinion are:
* To avoid an implicit dependency between hokusai and Datadog
* To avoid having to start injecting environment variables into all deployments
* Why limit the use of this value to one tool? E.g., some apps may want to print their current release identifier within a UI.

That's why I think it might be preferable for hokusai to just set a metadata label (either the common `app.kubernetes.io/version` or a custom one like `hokusai/deployment-digest`). That's already within the scope of its responsibilities (managing the spec), and Datadog or other tools can still be passed the value declaratively, like:

```yaml
name: DD_VERSION
valueFrom:
  fieldRef:
    fieldPath: metadata.labels['hokusai/deployment-digest']
```
This is analogous to how we already set the Datadog host to `spec.nodeName`, though it would require PRs to a few dozen projects.

artsyjian:
> ...The major reason _to do_ this would appear to be that all of our apps would get release-tracking with no further code/config changes. (Granted, this is a compelling one!)

Yep, I thought that was a requirement. Good that it's not since defaulting to `DD_VERSION` is very Artsy-specific. Okay, shall revise.

hokusai/services/deployment.py
```diff
@@ -88,6 +93,7 @@ def update(self, tag, constraint, git_remote, timeout, update_config=False, file
           for container in item['spec']['template']['spec']['containers']:
             if self.ecr.project_repo in container['image']:
               container['image'] = "%s@%s" % (self.ecr.project_repo, digest)
+              container['env'] = add_env(container, {'name': "%s" % app_version_var_name, 'value': "%s" % digest})
```
joeyAghion, line 96: Is this `digest` the whole long thing starting in `sha256:77af4d6b...`?

I don't remember exactly what the Datadog UI does, but maybe we should truncate this or strip the leading `sha256:`. (Ideally we'd have something more readable like the timestamps, but I don't think we have a "permanent" timestamp that spans environments handy.)

artsyjian: yeah stripping `sha256:` should be fine. timestamp could be useful, but wonder if it'd be confusing if the same image is deployed multiple times. i actually wanted to use git commit sha but refrained seeing how those tags in ECR are missing in git commit history (because of our squash-merge?)

artsyjian:
revised to use label with value of 2nd half of the image digest.

joeyAghion (approved):
2nd-half initially struck me as stranger than first-half (like git commits are often truncated), but I get it, given the common `sha256:` prefix.

If the same image gets deployed multiple times (e.g., a rollback), I actually think it's _good_ that the releases will match. In a rollback situation, the chart colors would accurately reflect the same release on either side of the broken release.

Ship it!

Issue #297, opened 2022-04-24
"hokusai review_app" command became "hokusai review-app" in v1.0.0


artsyjian:
The underscore became a dash, likely caused by dependency upgrades.

artsyjian:
https://github.com/pallets/click/issues/1123

PR #298, opened 2022-04-25
fix: click replacing review_app command with review-app

artsyjian:
Resolves https://github.com/artsy/hokusai/issues/297

Force Click to keep the underscore in `review_app` command.


ovasdi (approved):
LGTM

PR #299, opened 2022-04-26
feat: prepare version v1.0.1

artsyjian:
Prepare to release v1.0.1

### Notable changes since v1.0.0

08df37c fix: click replacing review_app command with review-app (#298)


ovasdi (approved):
Thanks for the fix!
🦖

PR #302, opened 2022-05-03
fix homebrew formula

artsyjian:
Ammend https://github.com/artsy/hokusai/pull/301

- use `bin.install` in homebrew formula, so that `brew install` creates `/usr/local/bin/hokusai` symlink
- let the build explicitly `git add` the formula file, so that if the file is not yet tracked by git, it will be


ovasdi (approved):
🦖

PR #319, opened 2022-11-23
docs(getting-started): Update with link to CircleCI setup instructions

damassi:
Missed this step, so thought it would helpful to update getting started docs with link to [further setup](https://github.com/artsy/README/blob/main/playbooks/deployments.md#setting-up-a-project-in-circleci).


araujobarret (approved):
Good link!

PR #326, opened 2023-01-23
fix: ci reference to dockerhub username

artsyjian:
The [previous PR](https://github.com/artsy/hokusai/pull/325) switched CI to use CircleCI context where the var for dockerhub username is configured as `DOCKERHUB_USER`. Update CircleCI config accordingly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,7 +146,7 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker
-      - run: echo "$DOCKERHUB_PASSWORD" | docker login --username "$DOCKERHUB_USER" --password-stdin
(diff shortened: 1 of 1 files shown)


joeyAghion (approved):

ovasdi:
@artsyjian 

As I already had a branch open what do you think of using this instead? https://github.com/artsy/hokusai/pull/329

It adds docker auth where images are pulled and switches to using `docker` context? Bringing us closer to consolidating dockerhub access to one context.

artsyjian:
Am informed of CircleCI's `docker` context which has `DOCKERHUB_USERNAME`, so updated PR to pull in that context as well, and revert to that var name.

Issue #328, opened 2023-01-23
commands fail when version is beta


artsyjian:
Hokusai commands fail when [version is beta](https://peps.python.org/pep-0440/#pre-releases).

Local Beta version of Hokusai installed in [editable mode](https://github.com/artsy/hokusai#poetry).
```
(hokusai) jian@artsy:~/code/hokusai$ git diff hokusai/VERSION
diff --git a/hokusai/VERSION b/hokusai/VERSION
index 21e8796..f1868ee 100644
--- a/hokusai/VERSION
+++ b/hokusai/VERSION
@@ -1 +1 @@
-1.0.3
+1.0.4b1

(hokusai) jian@artsy:~/code/hokusai$ hokusai version
1.0.4b1

(hokusai) jian@artsy:~/code/hokusai$ pyenv which hokusai
/home/jian/.pyenv/versions/hokusai/bin/hokusai
```

Running Beta version on a test project errors:
```
jian@artsy:~/code/hokusai-sandbox$ /home/jian/.pyenv/versions/hokusai/bin/hokusai check
ERROR: Hokusai's current version 1.0.4b1 does not satisfy this project's version requirements '>=0.5.8'.  Aborting.

jian@artsy:~/code/hokusai-sandbox$ /home/jian/.pyenv/versions/hokusai/bin/hokusai staging status
ERROR: Hokusai's current version 1.0.4b1 does not satisfy this project's version requirements '>=0.5.8'.  Aborting.
```

`1.0.4` is obviously >= `0.5.8`, so it seems the `b1` is causing a problem with the comparison.

artsyjian:
resolved by that pr

PR #330, opened 2023-01-31
feat: make the app aware when it is a beta version

artsyjian:
Currently `hokusai version` reads from [VERSION](https://github.com/artsy/hokusai/blob/main/hokusai/VERSION) file. This is true whether the installed Hokusai is from a [beta](https://github.com/artsy/hokusai#distributing-hokusai) or official release. This means when running `hokusai version` in a beta install, the output shows the last official release version, which is misleading and hinders testing.

This PR proposes to use [pre-release/dev version strings](https://peps.python.org/pep-0440/#version-scheme) for beta releases. Version strings for official releases remain the same.

The changes made to meet that goal:
- Use [setuptools_scm](https://github.com/pypa/setuptools_scm/) to guess the new version string. The string guessed is written to `hokusai/_version.py` (not checked into Git). `hokusai/version.py` is updated to read from that file. For example, given the current latest official release is `v1.0.3` (as indicated by Git tag), here's what is guessed on my local:
  ```
  $ python -m setuptools_scm 
  1.0.4.dev35+g7d06912

  $ cat hokusai/_version.py 
  # file generated by setuptools_scm
  # don't change, don't track in version control
  __version__ = version = '1.0.4.dev35+g7d06912'
  __version_tuple__ = version_tuple = (1, 0, 4, 'dev35', 'g7d06912')
  ```
- Update Makefile to run `setuptools_scm` command. Every `make` step that produces a distribution bundle (pyinstaller, docker, pip, ...) must run the command.
- Introduce `scripts/create_release_tag.sh`. This script must be run in every CircleCI `release` branch build. It reads `RELEASE_VERSION` file (replacement of the old `VERSION` file) which should be humanly bumped for every new official release. The script creates a Git tag (in CI) with the version string read, which in informs `setuptools_scm` to write that version to `hokusai/_version.py`. Without the Git tag, `setuptools_scm` writes a pre-release version which would be wrong for an official release. This is essentially human telling `setuptools_scm` what version string to use for official releases, as there's no way the tool can know which of Major/Minor/Patch version number we would like to bump.
- Switch docker build CI steps to use Python image (and install docker-cli) because `setuptools_scm` is now required.
- Update `check_required_version`:
  -  Make it work with pre-release version strings. Please see https://github.com/artsy/hokusai/issues/328 which this PR will resolve.
  - Add tests for pre-release/dev version strings.

Miscellaneous changes:
- Makefile:
  - Remove orphaned steps.
  - Rename steps for clarity.
  - Rename VERSION var to make clear the distinction between artifact label vs app version. Artifact label is used in names of distribution bundles, while app version is what's in `hokusai/_version.py`. Artifact label is either `latest`, `vX.Y.Z`, or `beta`, depending on type of release (official, beta). App version is always the actual version string determined by `setuptools_scm`.
- Re-order package list in `pyproject.toml` in alpha order.
- Make all `check_required_version` tests assert True, to make reading easier.


ovasdi (approved):
👏, a great improvement to the release process.

Issue #327, opened 2023-01-23
use 'main' as default for check-branch params


artsyjian:
[Current default](https://github.com/artsy/hokusai/blob/d6c6491d13356267a04dd8029a0da590c6c7a8ef/hokusai/commands/kubernetes.py#L50) is `master`. Many organizations, including Artsy, have moved on to `main`.

```
jian@artsy:~/code/hokusai-sandbox$ git rev-parse --abbrev-ref HEAD
main
jian@artsy:~/code/hokusai-sandbox$ hokusai staging update
ERROR: Not on master
 branch!  Aborting.
```

PR #345, opened 2023-03-24
feat: use 'main' as default for check-branch params [#FF]

artsyjian:
Solve https://github.com/artsy/hokusai/issues/327


hokusai/cli/production.py
```diff
@@ -4,7 +4,6 @@
 
 from hokusai.cli.base import base
 from hokusai.lib.common import set_verbosity, CONTEXT_SETTINGS
-from hokusai.lib.branch_detection import detect_branch
```
artsyjian, line 7: no use for this lib anymore now that Artsy has migrated to `main`.

ovasdi (approved):
🙌

PR #346, opened 2023-03-27
fix: push command not interpreting --force flag properly [#FF]

artsyjian:
`hokusai registry push` has a [`--force` option](https://github.com/artsy/hokusai/blob/d5f2cd3f5125c90041c5ea3a897d8e9245994200/hokusai/cli/registry.py#L19) for skipping Git status check. The option's `is_flag` is `True`, meaning if `--force` is not specified, it defaults to `False`. If `--force` is specified, it's value is `True`. So [`push`'s `force` param](https://github.com/artsy/hokusai/blob/d5f2cd3f5125c90041c5ea3a897d8e9245994200/hokusai/cli/registry.py#L23) always has value: `True` or `False`, never `None`.

The param is passed down to [`push` command](https://github.com/artsy/hokusai/blob/d5f2cd3f5125c90041c5ea3a897d8e9245994200/hokusai/commands/push.py#L11) which interprets it incorrectly - it tests `force` for `None`. This makes Git status checks always skipped.

There is a [similar error](https://github.com/artsy/hokusai/blob/d5f2cd3f5125c90041c5ea3a897d8e9245994200/hokusai/commands/push.py#L26) with the [`--overwrite` option](https://github.com/artsy/hokusai/blob/d5f2cd3f5125c90041c5ea3a897d8e9245994200/hokusai/cli/registry.py#L20).

This PR is to fix those errors, so that going forward, `--force` will be required to skip Git status check, and `--overwrite` will be required to overwrite tag in ECR.

Secondly, the Git status checks are a bit misleading. [`git status --porcelain --ignored`](https://github.com/artsy/hokusai/blob/d5f2cd3f5125c90041c5ea3a897d8e9245994200/hokusai/commands/push.py#L15) doesn't just list Git ignored files, it also lists untracked or modified files (as does `git status --porcelain`). If there are only untracked/modified files (no Git ignored files), `git status --porcelain --ignored` will say that there are Git ignored files, which is not true. this PR proposes to run `git status --procelain --ignored` only, and just show user the file list. And in the output, be more explicit about the [motivation](https://github.com/artsy/hokusai/blob/main/docs/Getting_Started.md) (item # 5) behind this check, which is to avoid copying unused/sensitive files into the Docker image, which I think should help user decide whether it's safe to force.

Users who have been using `hokusai registry push` locally likely have not used `--force` and `--overwrite`, so they will start encountering the output. I imagine they will switch to running the command with those flags all the time, but at least they will have been informed of the danger of doing so. I think they primarily use the command for creating Review Apps. [Force's review app script](https://github.com/artsy/force/blob/e3bee1f06162ffea072d6e8a215d6934331405b1/scripts/build_review_app.sh#L38) does use `--force` and `--overwrite`, so no change required there.

Lastly, I found the `push` command (one function with 7 `if`s) a bit hard to follow, so I refactored it and added tests to ensure it still behaves correctly.

### DIFF
--- a/hokusai/commands/push.py
+++ b/hokusai/commands/push.py
@@ -1,5 +1,4 @@
 import os
-import pdb
 
 from hokusai.lib.command import command
 from hokusai.lib.config import config
@@ -8,10 +7,9 @@ from hokusai.lib.common import print_green, print_red, shout
 from hokusai.lib.exceptions import HokusaiError
 from hokusai.services.docker import Docker
 
-@command()
 def build_and_push(remote_tag, local_tag, filename, ecr, skip_latest=False):
   Docker().build(filename)
-  push(remote_tag, local_tag, ecr, skip_latest)
+  push_only(remote_tag, local_tag, ecr, skip_latest)
 
 def ecr_check():
   ecr = ECR()
@@ -19,7 +17,6 @@ def ecr_check():
     raise HokusaiError("ECR repo %s does not exist... did you run `hokusai setup` for this project?" % config.project_name)
   return ecr
 
-@command()
 def git_status_check():
   git_unclean_files = shout('git status --porcelain --ignored')
   if git_unclean_files:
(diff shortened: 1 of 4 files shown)


hokusai/commands/push.py
```diff
@@ -1,42 +1,60 @@
...
+  """Check working directory's Git status"""
+  git_status_files = shout('git status --porcelain --ignored')
+  if git_status_files:
+    print_yellow("The following files/directories are either changed, untracked, or Git ignored.")
+    print_yellow("If they are not excluded by .dockerignore, they may get copied into the resulting Docker image.")
+    print_yellow(git_status_files)
+    raise HokusaiError("Working directory unclean. Aborting. Use '--force' to force.")
 
+@command()
+def push_image(remote_tag, local_tag, build, filename, force, overwrite, skip_latest=False):
```
artsyjian, line 30: `push.py` file is interpreted as a module by Python with module name `push`. When this function is named `push` and [imported in __init__py](https://github.com/artsy/hokusai/blob/d5f2cd3f5125c90041c5ea3a897d8e9245994200/hokusai/commands/__init__.py#L13), the function masks the module, so that the module cannot be imported elsewhere. Well, for the tests I have to [import the module](https://github.com/artsy/hokusai/pull/346/files#diff-796b61f71f490c9a3dd9fd895cc4133a85136b4c620a6a1102adb60da0fbb297R3), so renaming the function to `push_image`.

ovasdi (approved):
🙌

PR #353, opened 2023-04-27
feat: bump Docker Compose version to v2.17.2

artsyjian:
[PLATFORM-5143]

In order to support Docker multi-stage build, specifically the [`target` field in a `docker-compose.yml`](https://docs.docker.com/compose/compose-file/compose-file-v3/#target). That doc seems to say Compose spec must be versioned 3.4 or up for the field to work. In my testing, spec with version 2 also works, as long as `docker-compose` CLI understands the field.

The current version [does not understand the field](https://app.circleci.com/pipelines/github/artsy/exchange/7430/workflows/c657cf83-a96c-4792-945e-a43ae6f9b247/jobs/12202).

[PLATFORM-5143]: https://artsyproduct.atlassian.net/browse/PLATFORM-5143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN apk add --no-cache \
       openssl-dev
 
 # Install Docker Compose, AWS CLI
-RUN pip --no-cache-dir install docker-compose==1.25.5 && \
+RUN pip --no-cache-dir install docker-compose==2.17.2 && \
       pip --no-cache-dir install awscli --upgrade
 
 # Install the AWS IAM Authenticator
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 FROM python:3.9.10-alpine
 
+ARG COMPOSE_VERSION=2.17.2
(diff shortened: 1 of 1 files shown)


cavvia:
Nice... I wasn't familiar with the [multi-stage build](https://docs.docker.com/build/building/multi-stage/) feature but it looks powerful. In some repos I have a separate `Dockerfile.dev` but it seems that these can be unified with the prod image.

artsyjian:
> In some repos I have a separate `Dockerfile.dev`

That works too! It remains to be seen how easy/difficult it'd be to integrate multi-stage build into our Hokusai-managed projects. It was implemented in [Force](https://github.com/artsy/force/blob/main/Dockerfile) a while back. I hope the structure there can be applied to other projects.

PR #361, opened 2023-07-31
chore: bump Xcode test version to 13.4.1

mc-jones:
### Issue
As of August 7th, 2023, CircleCi will no longer host macOS images using Xcode `13.2.1`. As a result, ci will fail unless we update our config to specify a supported Xcode version.

[CircleCi Xcode image release schedule](https://circleci.com/docs/xcode-policy/?mkt_tok=NDg1LVpNSC02MjYAAAGMfanQyARdDkXPHZsqdicAiKVIH8hOhXNUlAd8ifo72Bjv1PpbC5xjHQfAon-AFjLjQIEwbCvGULm_KqogcGM_swyXoaTnlBrn7gxQrwxIOo_rbw)

### Solution
This PR updates the `test_macos` step of ci to use Xcode `13.4.1`.

[PHIRE-70]

[PHIRE-70]: https://artsyproduct.atlassian.net/browse/PHIRE-70?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pyproject.toml
```diff
@@ -20,16 +20,16 @@ Jinja2 = "==3.1.1"
...
+PyYAML = "==6.0.1"
 termcolor = "==1.1.0"
 
 [tool.poetry.dev-dependencies]
 coverage = "^6.3.2"
 httpretty = "^1.1.4"
 ipdb = "^0.13.9"
 mock = "^4.0.3"
-pyinstaller = "^4.5.1"
+pyinstaller = "^4.6"
```
mc-jones, line 32: `pyinstaller` needed to be bumped due to a [known issue](https://github.com/pyinstaller/pyinstaller/issues/6325) with a failing sanity check.

pyproject.toml
```diff
@@ -20,16 +20,16 @@ Jinja2 = "==3.1.1"
 MarkupSafe = "==2.1.1"
 packaging = "==21.3"
 prompt-toolkit = "==3.0.28"
-python = "^3.7"
-PyYAML = "==6.0"
+python = "==3.9.10"
```
mc-jones, line 23: With the previous version, `pyinstaller` was failing to resolve when running `poetry lock` due to the more permissive `python` version declared here. 

Additionally, it seems to make sense to pin to `3.9.10` since this is what `hokusai` is [tested on and the runtime version we explicitly declare](https://github.com/search?q=repo%3Aartsy%2Fhokusai%203.9.10&type=code).

pyproject.toml
```diff
@@ -20,16 +20,16 @@ Jinja2 = "==3.1.1"
...
+PyYAML = "==6.0.1"
```
mc-jones, line 24: I was seeing [test failures](https://app.circleci.com/pipelines/github/artsy/hokusai/843/workflows/1d1640d0-1db0-49ab-83be-34f3dd3c5869/jobs/3612?invite=true#step-105-214) resolving `pyYAML = "==6.0"` that were fixed with this `patch` version bump.

mc-jones:
To update the dependencies, I simply ran `poetry lock`. If I'm failing to understand any potential issues with how `python` dependencies resolve, please let me know.

joeyAghion (approved):

artsyjian:
Nice!

pyproject.toml
joeyAghion: This is causing problems [here](https://app.circleci.com/pipelines/github/artsy/horizon/2884/workflows/10e9a820-6eb7-4589-b4ea-a58af3878099/jobs/3400) because the _precise_ python version isn't available in alpine. What if we switched to something like `^3.9`?

artsyjian: let's do it (or even `^3.8` if we want to be safe). and watch for whether the `pyinstaller` problem recurs.

artsyjian: https://github.com/artsy/hokusai/pull/366

PR #370, opened 2023-08-18
feat: label review app namespaces

artsyjian:
Supports [PHIRE-179]

We [have a script](https://github.com/artsy/opstools/blob/7ca1362d01dff84f0f1b6b73504d51cef539375b/src/kubernetes_cleanup_namespaces/kubernetes_cleanup_namespaces/config.py#L25) that cleans up old Review Apps. It does so by deleting their Kubernetes namespaces.

The script must know which namespaces are Review Apps. Currently there is no way to determine that from the name of a namespace. Therefore, the script is hardcoded with a list of non-Review-App namespaces and it deletes namespaces not on that list. 

However, the list has become a burden to upkeep. This PR proposes to label Review App namespaces with `app-phase=review`. This allows the script to filter namespaces by that label, then we can do away with the non-Review-App list.

Un-related fix:
- Instruct to run `setuptools_scm` for local dev.

[PHIRE-179]: https://artsyproduct.atlassian.net/browse/PHIRE-179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


README.md
```diff
@@ -195,6 +195,14 @@ To update dependencies:
...
 ```
 
+### Local Dev Hokusai Version
+
+The version reported by `hokusai version` is taken from `hokusai/_version.py` file. The file is dynamically generated by [setuptools_scm](https://github.com/pypa/setuptools_scm) during CI and it is not version controlled. To have the file in local dev, run:
+
+```
+python -m setuptools_scm
+```
+
```
artsyjian, line 205: if this command is not run, `hokusai version` reports v0.0.0 (the [dummy string in pyproject.toml](https://github.com/artsy/hokusai/blob/c923cc674220b45841e5c186575ff95ac1b4d28a/pyproject.toml#L6)). that causes conflict with projects that require certain versions of Hokusai (e.g. [hokusai-sandbox](https://github.com/artsy/hokusai-sandbox/blob/40d5d97e7312900ce10a035c21ac3c8d2cc38b73/hokusai/config.yml#L4)).

hokusai/commands/namespace.py
```diff
@@ -25,7 +25,10 @@ def create_new_app_yaml(source_file, app_name):
       ('apiVersion', 'v1'),
       ('kind', 'Namespace'),
       ('metadata', {
-        'name': clean_string(app_name)
+        'name': clean_string(app_name),
+        'labels': {
+          'app-phase': 'review'
```
artsyjian, line 30: potentially another use case for a `~/.hokusai.conf` config file. i would have a var there like `org=acme`, then this label can become `acme/app-phase: review` which should never conflict with any label k8s might automatically add to this resource in the future.

mc-jones (approved):
This LGTM though, admittedly, I'm not familiar with the test patterns. I'll hold on the merge until in the event there is something that you want to follow up on after.

artsyjian:
thanks it should be good to merge.

Issue #372, opened 2023-08-25
Feature Request: a user Hokusai config file


artsyjian:
It would be great if Hokusai supports a user config file (e.g. `~/.hokusai.conf`). Potential use cases:

- Allows user to [specify a local `install-to` directory for `kubectl`](https://github.com/artsy/hokusai/issues/349#issuecomment-1693407419), so that `hokusai confgure` and kubectl-invoking code can reference this directory.

- Allows user to specify his/her org name (e.g. `artsy`) which Hokusai can use as a [prefix to Review App Kubernetes namespace labels](https://github.com/artsy/hokusai/pull/370#discussion_r1298863847).

artsyjian:
Turns out Hokusai [already supports a global config file](https://github.com/artsy/hokusai/blob/2b5f9b2a1dd97f7f7ed8a4e0a28e82d4681d682d/hokusai/lib/global_config.py#L16) at `~/.hokusai/config.yml`.

PR #376, opened 2023-09-12
feat: CommandRunner to load project's k8s secret if it exists

artsyjian:
[PHIRE-224]

- CommandRunner#run to load project's k8s secret if it exists. It does so by inserting a `secretRef` field when generating k8s container spec. The k8s secret may or may not exist, therefore, `optional: True` is also specified. CommandRunner#run is downstream to `hokusai staging/production/review_app run ...`, so with this in place, we will be able to [migrate a project's secrets from k8s configmap to Vault](https://artsyproduct.atlassian.net/browse/PHIRE-170?focusedCommentId=58698).

Related improvements:
- Currently, CommandRunner#run performs a bunch of tasks un-related to each other, and there are no unit tests, which makes updating scary. So:
  - Break down the function.
  - Move the following functions to lib, since they are not closely related to CommandRunner:
    - Obtaining user name from Env.
    - Validating key/value of Env var and nodeSelector constraint.
  - Add unit tests in Pytest format.

[PHIRE-224]: https://artsyproduct.atlassian.net/browse/PHIRE-224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hokusai/lib/common.py
+++ b/hokusai/lib/common.py
@@ -154,9 +154,9 @@ def user():
     )
   return user
 
-def validate_env_var(var):
-  ''' raise if env var is NOT of the form KEY=VALUE '''
-  if '=' not in var:
+def validate_key_value(key_value):
+  ''' raise if key_value is NOT of the form KEY=VALUE '''
+  if '=' not in key_value:
     raise HokusaiError(
-      "Error: environment variables must be of the form 'KEY=VALUE'"
+      "Error: key/value pair must be of the form 'KEY=VALUE'"
     )
--- a/hokusai/lib/common.py
+++ b/hokusai/lib/common.py
@@ -143,6 +143,17 @@ def pick_yes():
 def pick_no():
   return random.choice(["Nope", "No", "нет", "Ne", "नहीं", "Daabi", "Nein", "Nay", "Nē", "ні", "خیر", "Nie", "Non", "ניט", "не", "아니", "いや", "没有", "Não"])
 
+def user():
+  ''' obtain user name from environment '''
+  user = None
+  if os.environ.get('USER') is not None:
+    # The regex used for the validation of name is
+    # '[a-z0-9]([-a-z0-9]*[a-z0-9])?'
+    user = re.sub(
+      "[^0-9a-z]+", "-", os.environ.get('USER').lower()
+    )
+  return user
+
 def validate_env_var(var):
   ''' raise if env var is NOT of the form KEY=VALUE '''
   if '=' not in var:
--- a/hokusai/lib/common.py
+++ b/hokusai/lib/common.py
@@ -17,7 +17,7 @@ from botocore import session as botosession
 from termcolor import cprint
 
 from hokusai.lib.config import config
-from hokusai.lib.exceptions import CalledProcessError
+from hokusai.lib.exceptions import CalledProcessError, HokusaiError
 
 CONTEXT_SETTINGS = {
   'terminal_width': 10000,
@@ -142,3 +142,10 @@ def pick_yes():
 
 def pick_no():
   return random.choice(["Nope", "No", "нет", "Ne", "नहीं", "Daabi", "Nein", "Nay", "Nē", "ні", "خیر", "Nie", "Non", "ניט", "не", "아니", "いや", "没有", "Não"])
+
+def validate_env_var(var):
+  ''' raise if env var is NOT of the form KEY=VALUE '''
+  if '=' not in var:
+    raise HokusaiError(
+      "Error: environment variables must be of the form 'KEY=VALUE'"
+    )
--- a/hokusai/services/command_runner.py
+++ b/hokusai/services/command_runner.py
@@ -52,7 +52,8 @@ class CommandRunner:
 
   def _append_envfrom(self, container_spec):
     ''' append envFrrom to given container spec '''
-    container_spec = container_spec.update(
+    spec = copy.deepcopy(container_spec)
+    spec.update(
       {
         'envFrom': [
           {
@@ -69,6 +70,7 @@ class CommandRunner:
         ]
       }
     )
+    return spec
 
   def _append_constraints(self, containers_spec, constraint):
     ''' append constraints to given containers spec '''
--- a/hokusai/services/command_runner.py
+++ b/hokusai/services/command_runner.py
@@ -99,7 +99,7 @@ class CommandRunner:
     spec = self._set_envfrom(spec)
     return spec
 
-  def _overrrides_containers(self, cmd, env, tag_or_digest):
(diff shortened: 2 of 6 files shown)


hokusai/services/command_runner.py
```diff
@@ -1,71 +1,177 @@
...
+      {
+        'envFrom': [
+          {
+            'configMapRef': {
+              'name': f'{config.project_name}-environment'
+            }
+          },
+          {
+            'secretRef': {
+              'name': f'{config.project_name}',
```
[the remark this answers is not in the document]
artsyjian: b/c [the doc says to do that](https://www.notion.so/artsy/Hashicorp-Vault-77d94af51f714d51bb44049f4f2027bc?pvs=4#92489a9a9b34425aa08495b0d8623445).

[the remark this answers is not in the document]
cavvia: Right I was wondering about this convention

hokusai/services/command_runner.py
```diff
@@ -1,71 +1,177 @@
...
-    container = {
-      "args": cmd.split(' '),
-      "name": name,
-      "image": image_name,
-      "imagePullPolicy": "Always",
-      'envFrom': [{'configMapRef': {'name': "%s-environment" % config.project_name}}]
-    }
+  def _set_env(self, container_spec, env):
+    ''' set env field of given container spec '''
+    spec = copy.deepcopy(container_spec)
```
cavvia, line 46: Type hints would be useful here... it looks like `container_spec` is a dict, in which case I'm wondering why you are deepcopying it. Can you not just mutate it and return the mutated dict.

hokusai/services/command_runner.py
```diff
@@ -1,71 +1,177 @@
...
-      for s in env:
-        if '=' not in s:
-          raise HokusaiError("Error: environment variables must be of the form 'KEY=VALUE'")
-        split = s.split('=', 1)
-        container['env'].append({'name': split[0], 'value': split[1]})
-
-    spec = { "containers": [container] }
-    constraints = constraint or config.run_constraints
-    if constraints:
+  def _set_constraint(self, containers_spec, constraint):
```
cavvia, line 78: Type hints on the args here would again be useful

hokusai/services/command_runner.py
```diff
@@ -1,71 +1,177 @@
...
+    ''' set nodeSelector field of given containers spec '''
+    spec = copy.deepcopy(containers_spec)
```
cavvia, line 80: Again not convinced you need to deepcopy the arg if you are just setting it based on the return value of this function in line 117

hokusai/services/command_runner.py
```diff
@@ -1,71 +1,177 @@
...
+      'args': cmd.split(' '),
+      'name': self.container_name,
+      'image': self._image_name(tag_or_digest),
+      'imagePullPolicy': 'Always',
+    }
+    spec = self._set_env(spec, env)
+    spec = self._set_envfrom(spec)
+    return spec
+
+  def _overrrides_containers(self, cmd, env, tag_or_digest):
```
cavvia, line 102: There's a typo here (extra r in overrrides) but I also don't think you need this function.

hokusai/services/command_runner.py
```diff
@@ -1,71 +1,177 @@
...
+    spec = [container_spec]
+    return spec
+
+  def _overrides_spec(self, cmd, constraint, env, tag_or_digest):
+    ''' generate overrides['spec'] spec '''
+    spec = {}
+    containers_spec = self._overrrides_containers(
+      cmd, env, tag_or_digest
+    )
+    spec.update({'containers': containers_spec})
```
cavvia, line 116:
```suggestion
    containers_spec = self._overrides_container(
      cmd, env, tag_or_digest
    )
    spec.update({'containers': [containers_spec]})
```

cavvia (changes requested):
Good use of the pytest mock library here. Interesting also to see your nested function approach for declaring unit tests, which I think provides well structured output on the whole. We generally have been using unittest mock utilities over in ember for our testing but this looks pretty good too.

I have one potential blocker in terms of a method which seems superfluous to me under the current design.

hokusai/services/command_runner.py
```diff
@@ -1,71 +1,177 @@
-import os
-import re
+import copy
 import json
 import pipes
+import re
+
+import pdb
```
cavvia, line 6: Looks like dev code!

```suggestion
```

hokusai/lib/common.py
```diff
@@ -142,3 +142,21 @@ def pick_yes():
 
 def pick_no():
   return random.choice(["Nope", "No", "нет", "Ne", "नहीं", "Daabi", "Nein", "Nay", "Nē", "ні", "خیر", "Nie", "Non", "ניט", "не", "아니", "いや", "没有", "Não"])
```
cavvia, line 144: Very cosmopolitan!

hokusai/services/command_runner.py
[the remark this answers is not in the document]
artsyjian: ooo, making it configurable may be another use case for [the user config file that i am looking to add soon](https://github.com/artsy/hokusai/issues/372).

hokusai/services/command_runner.py
artsyjian: good catch on the typo. the tests didn't catch this because they have the typo too!

hokusai/services/command_runner.py
artsyjian: agreed

hokusai/services/command_runner.py
artsyjian: removed function

hokusai/services/command_runner.py
artsyjian: good catch!

hokusai/lib/common.py
artsyjian: oh yeah

artsyjian:
> We generally have been using unittest mock utilities over in ember...

oh nice that Pytest is used in Ember as well.

> ...a method which seems superfluous to me under the current design.

removed `_overrides_containers` if that's the method

hokusai/services/command_runner.py
artsyjian: > Can you not just mutate it...

Yeah, I prefer not to mutate args, when at all possible. The `deepcopy` is overhead, but the dict is not too large, so shouldn't hurt performance too much.

artsyjian:
About type hints, yeah, I have noticed them in Zenith repo. I wonder about introduction it in this repo...

cavvia (approved):

Issue #349, opened 2023-04-11
Docker Desktop's kubectl interferes with Hokusai's


artsyjian:
There has been cases where a Mac user's [PATH points to the `kubectl` [installed by Docker Desktop]](https://artsy.slack.com/archives/CP9P4KR35/p1681226052630619)(https://docs.docker.com/desktop/kubernetes/), and not to `/usr/local/bin/kubectl` (installed by Hokusai). So far we've seen this for M1 users.

In one case, Docker Desktop's `kubectl` lives in `~/.docker/...` and this dir is listed at the very beginning of user's PATH. The user's Shell init script (in this case, for `zsh`) has been modified by Docker Desktop to `source` a file which presumably did that to the PATH. The work-around for this specific case is to comment out the `source` or add a line in the Shell init script toward the bottom to undo that PATH change, but this is cumbersome and might not survive the next Docker Desktop install/update.

I am not sure what Hokusai can do to totally solve the problem, but some ideas for mitigation:

- Currently `hokusai check` simply checks that [`which kubectl` returns no error](https://github.com/artsy/hokusai/blob/89af1e9726af3e41ce46a61daf3b22132ac5d7a2/hokusai/commands/check.py#L46). We can have it read the returned path and warn if it's not `/usr/local/bin`, which can help user notice a path problem.

- Currently `hokusai configure` [copies the downloaded `kubectl` to `/usr/local/bin/kubectl`](https://github.com/artsy/hokusai/blob/89af1e9726af3e41ce46a61daf3b22132ac5d7a2/hokusai/commands/configure.py#L41). Perhaps it's feasible to name the executable differently in `/usr/local/bin`, for example, `hokusai_kubectl`, then there should no longer be any conflict with Docker Desktop. The problem is that if user ever has to run kubectl directly, the user must know about `hokusai_kubectl`.

artsyjian:
More ideas:

- `hokusai check` can check the version of the `kubectl` binary, wherever that binary is stored. If the version matches that of the Kubernetes cluster, then there's no problem, otherwise, fail the check. So we do not care about whether the `kubectl` binary was installed by Hokusai or Docker Desktop, we only care about the version.

[...]

PR #379, opened 2023-09-29
feat: add a user global config file

artsyjian:
Solves https://github.com/artsy/hokusai/issues/372, https://github.com/artsy/hokusai/issues/349

- Update list of global configs, now to be:
  - kubeconfig_dir
  - kubeconfig_source_uri
  - kubectl_dir
  - kubectl_version
- Update `lib/global_config.py`
  - Allow config file to be specified by HOKUSAI_GLOBAL_CONFIG env var (pointing to S3 or local file), default to `~/.hokusai.yml`
  - Add functions to HokusaiGlobalConfig class
    - `merge` - Allows overriding config
    - `save` - Allows saving config to `~/.hokusai.yml` (file path not configurable)
- Update `commands/configure.py`
  - Update command line options:
    - Remove the following as they are org-wide and can be read from global config
      - `bucket_name`, `key_name` (correspond with `kubeconfig_source_uri` global config)
      - `kubectl_version`
    - Rename the following, and keep them as they may differ from local to local
      - `install_to` (rename to `kubectl_dir`)
      - `install_config_to` (rename to `kubeconfig_dir`)
  - Allow skipping kubectl and kubeconfig installs
  - After reading global config, override it with user's customizations (i.e. user supplied command line options)
  - Save global config

The changes make possible the following:

- Org users bootstrap their `~/.hokusai.yml` from prepared org-wide config (e.g. in S3). They run `HOKUSAI_GLOBAL_CONFIG=s3://bucket/path/to/hokusai.yml hokusai configure`. User can add options to customize a limited number of configs, for example: `HOKUSAI_GLOBAL_CONFIG=s3://bucket/path/to/hokusai.yml hokusai configure --kubectl-dir /my/special/dir`

- To distribute new org-wide config (e.g. new `kubectl_version`), org admin updates `s3://bucket/path/to/hokusai.yml` or prepare a new file. Users run: `HOKUSAI_GLOBAL_CONFIG=... hokusai configure`

- `~/.hokusai.yml` is no longer read only, but read/write. Configs from S3 HOKUSAI_GLOBAL_CONFIG or `hokusai configure` options are persisted in `~/.hokusai.yml`, available to any hokusai command at any time. Indeed, as part of this PR, `services/kubectl.py` will read `kubectl_dir` config for path to kubectl executable. The [Docker-Desktop/Hokusai kubectl conflict problem](https://github.com/artsy/hokusai/issues/349) is resolved, if a goood kubectl dir is specified (e.g. `~/.local/bin`, and not `/usr/local/bin`), in org-wide config or by `hokusai configure --kubectl-dir` option.

- If user's kubectl is deleted accidentally, user can run `hokusai configure` (without any args) and restore it, because all necessary configs are available in `~/.hokusai.yml`. User can also customize any config by editing `~/.hokusai.yml` and run `hokusai configure` (without any args).

- User can run any Hokusai command (e.g. `hokusai staging status`) with HOKUSAI_GLOBAL_CONFIG env var pointing to a config file other than `~/.hokusai.yml`. I don't see a good use case for it, since that is reading a global config not saved by `hokusai configure`, and those configs might not be compatible with user's local (e.g. location of kubectl executable). This is a side effect of using an env var (due to [problem with using `--new-config` option](https://github.com/artsy/hokusai/pull/379#discussion_r1342888566)).

Related changes:
- Rename [configure function](https://github.com/artsy/hokusai/blob/09bcc2fc76249af077e6230c697d1232211ff2fa/hokusai/commands/configure.py#L19) to `hokusai_configure` to resolve Pytest errors. Currently, a lot of cases (e.g. hokusai/commands folder) where the same name is used for the module (the file) and a function inside it. Multiple times this has caused problem for testing.
- Retrofit tests for classes/functions touched (e.g. [Kubectl service](https://github.com/artsy/hokusai/blob/main/hokusai/services/kubectl.py)), to help spot regression. (Classes/functions added by this PR of course have tests added along, except for [configure command](https://github.com/artsy/hokusai/blob/09bcc2fc76249af077e6230c697d1232211ff2fa/hokusai/commands/configure.py#L18) which I cannot find a way to test due to it having a decorator.)
- Add `S3Interface` class for interfacing with S3 rather than having S3 client sprinkled here and there.
- Refactor the code quite a bit to make them easier to test, less redundant, more consistent, generic functions moved to lib, etc.
- Cosmetic changes: line ordering, line length, quoting consistency, etc.

Un-related changes:
- Fix [this SSL_CERT_FILE env assignment](https://github.com/artsy/hokusai/blob/09bcc2fc76249af077e6230c697d1232211ff2fa/bin/hokusai#L8). It causes problem for Pip/local-dev installed Hokusai, where the [lib function returns empty string](https://github.com/artsy/hokusai/blob/09bcc2fc76249af077e6230c697d1232211ff2fa/hokusai/lib/environment.py#L15), and when the env var is set to that, OpenSSL fails to find certs, breaking all HTTPS requests. This problem was introduced almost [1.5 years ago](https://github.com/artsy/hokusai/pull/301/files#diff-19fc065cebe5a02e23ef9a7bae7130d60d518e932ddb05d41441401cfe0795eaR6). A smoke test will have to be added to prevent this problem in the future.

Migration:
---
- [x] Create org-wide global configs in S3:
  - [x] s3://artsy-provisioning-public/hokusai/hokusai-ci.yml
    ```
    ---
    kubeconfig-dir: ~/.kube
    kubeconfig-source-uri: s3://artsy-citadel/k8s/config-ci
    kubectl-dir: /usr/local/bin
    kubectl-version: 1.19.16
    ```
  - [x] s3://artsy-provisioning-public/hokusai/hokusai-dev.yml
    ```
    ---
    kubeconfig-dir: ~/.kube
    kubeconfig-source-uri: s3://artsy-citadel/k8s/config-dev
    kubectl-dir: ~/.local/bin
    kubectl-version: 1.19.16
    ```
  - [x] s3://artsy-provisioning-public/hokusai/hokusai-admin.yml
    ```
    ---
    kubeconfig-dir: ~/.kube
    kubeconfig-source-uri: s3://artsy-citadel/k8s/config-admin
    kubectl-dir: ~/.local/bin
    kubectl-version: 1.19.16
    ```
- [x] Update [Hokusai orb](https://github.com/artsy/orbs/blob/7c26a69a6b560d6acb4b31fb267244d89ea42d6a/src/hokusai/hokusai.yml#L54C27-L54C52) to run `hokusai configure` in two different ways depending on Hokusai version.
    - For current Hokusai version, do:
      ```
      configure-hokusai:
        parameters:
          configUri:
            type: string
            default: "https://artsy-provisioning-public.s3.amazonaws.com/hokusai/config-ci.yml"
        steps:
          - run:
              name: Configure Hokusai
              command: |
                mkdir -p ~/.hokusai
                curl -o ~/.hokusai/config.yml << parameters.configUri >>
                hokusai configure
      ```
  - For new (released by this PR), do:
      ```
      configure-hokusai:
        steps:
          - run:
              name: Configure Hokusai
              command: |
                HOKUSAI_GLOBAL_CONFIG=s3://artsy-provisioning-public/hokusai/config-ci.yml hokusai configure
      ```
- [ ] Release the changes in this PR as Hokusai v2.0.0, bumping major version from 1 to 2.
- [ ] Delete old global config files on S3:
  - [ ] s3://artsy-provisioning-public/hokusai/config-ci.yml
  - [ ] s3://artsy-provisioning-public/hokusai/config-dev.yml
  - [ ] s3://artsy-provisioning-public/hokusai/config-admin.yml
- [ ] Update docs in other repos:
  - [ ] README
  - [ ] Potential
- [ ] Ask Sapphire team to run: `HOKUSAI_GLOBAL_CONFIG=s3://artsy-provisioning-public/hokusai/hokusai-dev.yml hokusai configure`
- [ ] Ask developers to run: `HOKUSAI_GLOBAL_CONFIG=s3://artsy-provisioning-public/hokusai/hokusai-dev.yml hokusai configure`
- [ ] Remove Hokusai v1 logic from [Hokusai orb](https://github.com/artsy/orbs/blob/7c26a69a6b560d6acb4b31fb267244d89ea42d6a/src/hokusai/hokusai.yml#L54C27-L54C52).


hokusai/cli/base.py
```diff
@@ -24,18 +24,16 @@ def console():
...
-@click.option('--install-to', type=click.STRING, default='/usr/local/bin', help='Install kubectl to (default: /usr/local/bin)')
-@click.option('--install-config-to', type=click.STRING, default=os.path.join(os.environ.get('HOME'), '.kube'), help='Install kubectl config to (default: ~/.kube)')
-@click.option('-v', '--verbose', type=click.BOOL, is_flag=True, help='Verbose output')
-def configure(kubectl_version, s3_bucket, s3_key, config_file, platform, install_to, install_config_to, verbose):
-  """Install and configure kubectl"""
+@click.option('--kubeconfig-dir', type=click.STRING, default=None, help='Directory to install kubeconfig into. The kubeconfig file will be named "config". (default: None)')
+@click.option('--kubectl-dir', type=click.STRING, default=None, help='Directory to install kubectl into. The executable will be named "kubectl". (default: None)')
+@click.option('--new-config', type=click.STRING, default=None, help='Path to new Hokusai global config. Can be AWS S3 location (e.g. s3://bucket/prefix/to/file), or local file (e.g. /path/to/file, ./path/to/file). New config will overwrite existing ~/.hokusai.yml')
```
joeyAghion, line 29: This naming is a little unclear to me. It's like a config "source" or "import," basically, right? I will browse some other CLIs' options to refine an opinion about this...

joeyAghion: Actually, having read through this more fully now, I think the closest analog is `kubeconfig-source-uri`. Why not name this `source-uri`? The case of importing from a local file could maybe still be supported, but that scenario seems odd to me so I'm curious how you think it would be used.

joeyAghion (commented):
Idea: the source config path could _itself_ be persisted in the configuration file, and then future calls to `hokusai configure` (with no args) could re-import from that automatically. What do you think?

README.md
```diff
@@ -216,19 +216,13 @@ minikube start --kubernetes-version=<version of your Kubernetes clusters, exampl
 
 ### Run tests
 
-To run all tests:
-
-```
-make tests
```
artsyjian, line 222: can no longer offer this. it runs unit and integrations tests in tandem but the 2 suites conflict on HOME env. so each suite must be run separately.

hokusai/cli/base.py
artsyjian: `source-uri` sounds ambiguous but the help string does elaborate `Path to new Hokusai global config...`, so I suppose it's fine. Allowing local file source is so that one is not locked into using S3.

artsyjian:
> Idea: the source config path could _itself_ be persisted in the configuration file, and then future calls to `hokusai configure` (with no args) could re-import from that automatically. What do you think?

Ah, so that means `hokusai configure ...` will always overwrite `~/.hokusai.yml`. It might be fine. Although that will rule out this use case which we might want to allow: user directly modifying `~/.hokusai.yml` and then run `hokusai configure` (with no arg).

Some configs (e.g. kubectl-dir) have corresponding command line options which user can use to re-customize. Other configs (e.g. kubectl-version) can be customized only by editing `~/.hokusai.yml`, though I am not sure why it should ever be done.

hokusai/lib/global_config.py
```diff
@@ -1,54 +1,69 @@
...
+    return self._config['kubeconfig-source-uri']
 
+  @property
+  def kubectl_dir(self):
+    return self._config['kubectl-dir']
 
-global_config = HokusaiGlobalConfig()
+  @property
+  def kubectl_version(self):
+    return self._config['kubectl-version']
```
artsyjian, line 69: wonder if this is better called 'k8s-version', looking at it from cluster's perspective, and kubectl should match cluster version anyway.

hokusai/cli/base.py
artsyjian: Alas, this command line option does not work. The way we [load all modules upfront](https://github.com/artsy/hokusai/blob/09bcc2fc76249af077e6230c697d1232211ff2fa/hokusai/__init__.py#L4) when running any Hokusai command, Kubectl service gets loaded too, and it [instantiates global config object](https://github.com/artsy/hokusai/pull/379/files#diff-a4ddf5adce9f1ad6808af8d4aa9dcaa680aebc2d4fa1ba58c453b245602f32ccR10) at module level. That instantiation is immediately done when the module is imported. The result is that when any Hokusai command is run, `~/.hokusai.yml` must already exist, so user won't get a chance to run `hokusai configure --new-config`.

This blanket upfront loading of modules has been causing problem again and again. I am not sure what to do about it at the moment.

For now, we can just [replace the command line option with an Env var](https://github.com/artsy/hokusai/pull/379/commits/60f16a31d33dd2de0c66521c80d2b8398334a01c) which takes precedence over `~/.hokusai.yml`, for config read. Config write will continue to go to `~/.hokusai.yml`.

artsyjian: The error I got was:
```
artsy:hokusai-sandbox jxu$ /Users/jxu/.pyenv/versions/hokusai/bin/hokusai configure --help
Error: failed to copy /Users/jxu/.hokusai.yml to /var/folders/n6/mjsp4zt94yn60gfn7dzr248w0000gn/T/tmpkd40shy_/hokusai.yml
Error: Not able to load config from /Users/jxu/.hokusai.yml
  File "/Users/jxu/.pyenv/versions/hokusai/bin/hokusai", line 3, in <module>
    from hokusai.command_line import main
  File "/Users/jxu/code/hokusai/hokusai/__init__.py", line 4, in <module>
    from hokusai.commands import *
  File "/Users/jxu/code/hokusai/hokusai/commands/__init__.py", line 2, in <module>
    from hokusai.commands.check import check
  File "/Users/jxu/code/hokusai/hokusai/commands/check.py", line 12, in <module>
    from hokusai.services.kubectl import Kubectl
  File "/Users/jxu/code/hokusai/hokusai/services/kubectl.py", line 10, in <module>
    global_config = HokusaiGlobalConfig()
  File "/Users/jxu/code/hokusai/hokusai/lib/global_config.py", line 15, in __init__
    self._config = ConfigLoader(uri).load()
  File "/Users/jxu/code/hokusai/hokusai/lib/config_loader.py", line 32, in load
    uri_to_local(self.uri, tmpdir, filename)
  File "/Users/jxu/code/hokusai/hokusai/lib/common.py", line 197, in uri_to_local
    local_to_local(parsed_uri.path, target_dir, target_file)
  File "/Users/jxu/code/hokusai/hokusai/lib/common.py", line 97, in local_to_local
    shutil.copyfile(source_path, target_path)
  File "/Users/jxu/.pyenv/versions/3.9.10/lib/python3.9/shutil.py", line 264, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/jxu/.hokusai.yml'
```

joeyAghion (approved):
> ...that will rule out this use case which we might want to allow: user directly modifying ~/.hokusai.yml and then run hokusai configure (with no arg).

This seems OK to me, since a `hokusai configure --some-arg VALUE` can always be performed to write _and remember_ a new setting, rather than needing to edit the config manually.

In any case, I _believe_ this latest version of the code is opting for a `HOKUSAI_GLOBAL_CONFIG` env var that is not itself persisted as a setting. That's fine too! It just means rolling out new configs to the org will require devs run `HOKUSAI GLOBAL_CONFIG=... hokusai configure` instead of `hokusai configure`.

Up to you how to proceed, but can we update the README or command reference doc accordingly? Sometimes explaining the behavior reveals how it can be improved.

artsyjian:
> but can we update the README or command reference doc accordingly? Sometimes explaining the behavior reveals how it can be improved.

yeah, i think i finally got the PR description worded right, and even that got me thinking hard about the changes. will work on the docs.

artsyjian:
updated README and docs under `doc` folder.

docs/hokusai_files.md
```diff
@@ -0,0 +1,200 @@
...
+
+The following Hokusai files must exist on a user's local environment:
+
+Global files:
+
+- `~/.hokusai.yml` - user's global Hokusai configuration
+
+Per-project files:
+
+These files are stored in a `hokuai` directory that is in the project's root directory (i.e. `/path/to/project/hokusai`)
```
joeyAghion, line 11: `hokusai`

README.md
```diff
@@ -82,31 +82,31 @@ Hokusai currently supports Python 3.7+ only. The last version that supported Pyt
...
+1. Ensure your org admin has completed the steps mentioned in [Administering Hokusai](./docs/Administering_Hokusai.md).
 
-2. Configure Hokusai
+2. Configure your local environment with [AWS credentials](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#configuring-credentials).
+
+3. Run Hokusai configure command:
 
     ```
-    hokusai configure --kubectl-version <kubectl version> --s3-bucket <bucket name> --s3-key <file key>
+    HOKUSAI_GLOBAL_CONFIG=<path-to-org-wide-global-config-file> hokusai configure
```
joeyAghion, line 94: It's a little unclear if `configure` can be run without a global config, or what defaults are imposed if so.

~Maybe an example config file could help, or an example of also passing args.~ Nevermind I see that's discussed below. (Still not sure if a global config is _required_ though.)

joeyAghion (approved):
Thanks and sorry if my suggestion triggered the more exhaustive rewrite!

docs/hokusai_files.md
artsyjian: 👍

README.md
artsyjian: [how now?](https://github.com/artsy/hokusai/pull/379/commits/8de9f317beca06b94608b8d2bb45f6d8fef459f3)

docs/Command_Reference.md
```diff
@@ -1,136 +1,174 @@
...
+5) Wait for the Deployment to roll out. If rollout does not complete within `--timeout` minutes (default 10), run `kubectl rollout undo` to roll all Deployments back automatically, then exit with `1`.
+6) If `post-deploy` hook command is specified, run it. If it fails, print a warning and continue.
+7) Push deployment Docker tags to the registry. If it fails, print a warning and continue.
+8) If `git-remote` is specified in `./hokusai/config.yml`, push deployment Git tags to Github. If it fails, print a warning and continue.
+9) Finally, exit `0` if all steps were successful. If any of steps `6`, `7` or `8` failed, exit with `1`.
 
-### How to do a rollback
 
-You can use the command `hokusai registry images` to get a list of all images for your current project. They order from most recent to oldest. To do a rollback use `hokusai production deploy [image_tag]` to get back to a particular version. 
+### How to rollback
```
artsyjian, line 159: this section seems more appropriate for the Hokusai playbook in README repo.

README.md
joeyAghion: 👍

artsyjian:
Orb side pr: https://github.com/artsy/orbs/pull/158

PR #382, opened 2023-10-09
fix: setuptools_scm not writing _version.py

artsyjian:
Discovered that `python -m setuptools_scm` no longer respects [write_to](https://github.com/artsy/hokusai/blob/00af3b57fad769d57cfac77c95ad7dc5c5e8172c/pyproject.toml#L50), that is, it doesn't create `hokusai/_version.py` file anymore. We have to do `poetry build` now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,6 +103,7 @@ jobs:
       - run: apk add curl docker-cli git make
       - run: pip install setuptools --upgrade
       - run: pip install setuptools_scm
+      - run: pip install poetry
       - run: make build-docker-image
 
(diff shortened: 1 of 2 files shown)


damassi (approved):

.circleci/config.yml
```diff
@@ -256,6 +258,7 @@ jobs:
       - run: apk add curl docker-cli git make
       - run: pip install setuptools --upgrade
       - run: pip install setuptools_scm
+      - run: pip install poetry
```
damassi, line 261: A bit unfortunate, but alas

Issue #384, opened 2023-10-10
Using setuptools_scm causes too much complexity and flake in CI builds


artsyjian:
https://github.com/artsy/hokusai/pull/330 introduced [setuptools_scm](https://github.com/pypa/setuptools_scm) for versioning Beta releases. It is used to version canonical releases as well, but as it doesn't obviate our hand editing a version file, it's a burden for that purpose, rather than a benefit.

Recently [setuptools_scm v8 was released](https://github.com/pypa/setuptools_scm/releases/tag/v8.0.0) which has breaking changes, and as our Hokusai Docker image builds [install the package outside of Poetry](https://github.com/artsy/hokusai/blob/b7fea74ce8dd3a2b4ae6a0d9cb4a8a5b9f507fa0/.circleci/config.yml#L105), the builds pick up the new version and consequently broke (i.e. [write_to doesn't work anymore](https://github.com/artsy/hokusai/pull/382), a [newer version of setuptools is required](https://github.com/artsy/hokusai/pull/381) but this constraint is not defined in setuptools_scm package).

That and the following adjustments that were required to integrate setuptools_scm:

- Incompatibility with Poetry in using `pyproject.toml` file. Poetry expects [version spec in that file](https://github.com/artsy/hokusai/blob/b7fea74ce8dd3a2b4ae6a0d9cb4a8a5b9f507fa0/pyproject.toml#L6) which currently holds a dummy version (`0.0.0`). That dummy version is interpreted by PyPI build so we actually have to edit the version string in pyproject.toml using [poetry version command](https://github.com/artsy/hokusai/blob/b7fea74ce8dd3a2b4ae6a0d9cb4a8a5b9f507fa0/Makefile#L100). 
- Sprinkling `python -m setuptools_scm` in [Makefile](https://github.com/artsy/hokusai/blob/b7fea74ce8dd3a2b4ae6a0d9cb4a8a5b9f507fa0/Makefile#L30).
- For Docker builds, since we don't want dev dependencies in the image (e.g. `poetry`, `setuptools_scm`), we must install setuptools_scm outside of Docker build, currently done in CircleCI config. For non-Docker builds, `poetry install` is used, for Docker builds, it's `pip install`, so the version of setuptools_scm used can be different between the two types of builds.
- [A custom script](https://github.com/artsy/hokusai/blob/main/scripts/create_release_tag.sh) is required to get setuptools_scm to emit canonical versions.
[...]

Issue #357, opened 2023-07-03
hokusai version command shows dev version, when installed via Pip


artsyjian:
Install Hokusai via Pip:

```
pip install hokusai
```

Hokusai version command shows a dev version:
```
$ hokusai version
1.1.1.dev0+gb6a5062.d20230419
```

`_version.py` (generated by setuptools_scm) indeed has dev version:
```
jian@artsy:~/.pyenv/versions/3.9.10/lib/python3.9/site-packages/hokusai$ cat _version.py
# file generated by setuptools_scm
# don't change, don't track in version control
__version__ = version = '1.1.1.dev0+gb6a5062.d20230419'
__version_tuple__ = version_tuple = (1, 1, 1, 'dev0', 'gb6a5062.d20230419')
```

The Makefile's [publish-to-pip step](https://github.com/artsy/hokusai/blob/main/Makefile#L96) runs `setuptools_scm` which generates the canonical version (1.1.0). The problem is with the subsequent commands:
- `poetry version`, which bumps the [version stored in pyproject.toml](https://github.com/artsy/hokusai/blob/d4a9714779298d6348ba25a188eb3f9ab6bd65c1/pyproject.toml#L6), to `1.1.0`.
- `poetry build`, which actually runs `setuptools_scm` too, and this time, because of the version bump in pyproject.toml, generates a dev version.

For example, if we were to release v1.2.0:
```
(hokusai) jian@artsy:~/code/hokusai$ python -m setuptools_scm
1.2.0

(hokusai) jian@artsy:~/code/hokusai$ cat hokusai/_version.py
# file generated by setuptools_scm
# don't change, don't track in version control
__version__ = version = '1.2.0'
__version_tuple__ = version_tuple = (1, 2, 0)

(hokusai) jian@artsy:~/code/hokusai$ poetry version 1.2.0
Bumping version from 0.0.0 to 1.2.0

(hokusai) jian@artsy:~/code/hokusai$ poetry build
Building hokusai (1.2.0)
  - Building sdist
  - Built hokusai-1.2.0.tar.gz
  - Building wheel
  - Built hokusai-1.2.0-py3-none-any.whl

(hokusai) jian@artsy:~/code/hokusai$ cat hokusai/_version.py
# file generated by setuptools_scm
# don't change, don't track in version control
__version__ = version = '1.2.1.dev0+gb6a5062.d20230703'
[...]

PR #389, opened 2023-12-27
feat: new way of versioning beta builds

artsyjian:
[setuptools_scm](https://github.com/artsy/hokusai/pull/330) was introduced around a year ago to help version Beta builds. [Our experience](https://github.com/artsy/hokusai/issues/384) shows that it is not worth the trouble.

This PR removes setuptools_scm, restores hokusai/VERSION file and much of the pre-setuptools_scm configs, and introduces a different way of versioning Beta builds.


hokusai/VERSION
```diff
@@ -0,0 +1 @@
+999.999.999
```
artsyjian, line 1: not 0.0.0 b/c [some artsy projects](https://github.com/artsy/gravity/blob/caf28fd0345e50ae928d66404e3b15fc0b902707/hokusai/config.yml#L5) have min Hokusai version requirements.

hokusai/lib/environment.py
```diff
@@ -26,3 +26,9 @@ def templates_dir_path():
     return sys._MEIPASS + '/hokusai_datas/templates'
   else:
     return os.path.dirname(hokusai.__file__) + '/templates'
+
+def version_file_path():
+  if frozen():
+    return sys._MEIPASS + '/hokusai_datas/VERSION'
+  else:
+    return os.path.dirname(hokusai.__file__) + '/VERSION'
```
artsyjian, line 34: hokusai/VERSION file path differs depending on the build, PyInstaller build vs. Pip vs. local

artsyjian:
Should resolve https://github.com/artsy/hokusai/issues/384 and https://github.com/artsy/hokusai/issues/357.

hokusai/VERSION
artsyjian: it's just a placeholder

mc-jones: Minor: could be worth adding this context as a comment as it is non-obvious.

mc-jones (approved):
Nice change as I've been bit by the versioning issue before. 👍

hokusai/VERSION
artsyjian: True. But a comment would complicate [the code that reads the file](https://github.com/artsy/hokusai/pull/389/files#diff-c289bc883a0cbdc59ef79744e976404261024f587ca9a8b3af08e1396da946a8R6). How about [adding the info to README](https://github.com/artsy/hokusai/pull/389/commits/e59d52944a88e7a82d62968e48e0488b623bfb28)?

mc-jones: That works!

Issue #126, opened 2018-12-19
List review apps


jonallured:
Unless I'm missing something, I don't think there's a way to see a list of the review apps for a project. This would be a nice way to ensure we're using the right name for an app and that we've cleaned up after ourselves.

izakp:
@jonallured `hokusai review_app setup` will create a Yaml file for that review app in the `./hokusai/` directory.  To the best of my recollection, @ashkan18 and I designed this so we can commit that file for the lifecycle of the review app.

jonallured:
Ok, so you're saying that YAML file should be checked in while the PR is under review in one commit and then later on you'd do another commit that removes it? Maybe we just need to document this workflow better?

artsyjian:
yeah, a `hokusai review_app list` command may be nice.

artsyjian:
review app namespaces are already labeled with `app-phase=review`:
```
$ kubectl --context staging get ns -l app-phase=review
NAME                      STATUS   AGE
alerts-button-expansion   Active   26d
inbox2                    Active   29d
```
That together with our `app` [Deployment label](https://github.com/artsy/force/blob/e2ab6f0b22752d9bdc2fb1722af986e392bd9898/hokusai/staging.yml#L26), Hokusai should be able to identify all review apps of a project.

PR #390, opened 2023-12-27
feat: hokusai review_app list command

artsyjian:
Solves https://github.com/artsy/hokusai/issues/126

Introduce `hokusai review_app list` command which lists review apps of the project. It does so by listing Staging cluster namespaces that are labeled with:

- app-name=<...project-name...>
- app-phase=review

`app-phase` label is already being applied to review app namespaces. This PR introduces `app-name` label.

Example usage:
```
artsy:force jxu$ /Users/jxu/.pyenv/versions/hokusai/bin/hokusai review_app list
inbox2
```


mc-jones (approved):
Nice feature!

PR #392, opened 2023-12-29
feat: un-decorate all commands, and print stacktrace for unanticipated exceptions

artsyjian:
The [CLI](https://github.com/artsy/hokusai/tree/main/hokusai/cli) layer invokes [commands](https://github.com/artsy/hokusai/tree/main/hokusai/commands). The commands invoked should always be decorated ([example](https://github.com/artsy/hokusai/blob/004ef7887a3b7fdc18277a92b114e427a70e6eeb/hokusai/commands/namespace.py#L18)) by [command lib function](https://github.com/artsy/hokusai/blob/004ef7887a3b7fdc18277a92b114e427a70e6eeb/hokusai/lib/command.py#L10), for proper exception handling, among other things.

But the decorator disrupts testing. For example, if [list_namespaces](https://github.com/artsy/hokusai/blob/004ef7887a3b7fdc18277a92b114e427a70e6eeb/hokusai/commands/namespace.py#L48) were decorated with `@command()`:

```
(hokusai) artsy:hokusai jxu$ pytest test/unit/test_commands/test_namespace.py 
...
test/unit/test_commands/test_namespace.py::describe_list_namespaces::it_lists FAILED...

    def it_lists(mocker, mock_kubectl_obj, capfd):
      mocker.patch('hokusai.commands.namespace.Kubectl', return_value=mock_kubectl_obj)
      spy = mocker.spy(mock_kubectl_obj, 'get_objects')
>     list_namespaces('staging', 'foo1=bar1')

test/unit/test_commands/test_namespace.py:12: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = ('staging', 'foo1=bar1'), kwargs = {}, result = None

    @wraps(func)
    def wrapper(*args, **kwargs):
      try:
        if config_check:
          config.check()
        result = func(*args, **kwargs)
        if result is None:
>         sys.exit(0)
E         SystemExit: 0

hokusai/lib/command.py:19: SystemExit
...
```

Test fails because it's the decorated `list_namespaces` being tested, not the original `list_namespaces`.

The existing solution is to [mock things done by the decorator](https://github.com/artsy/hokusai/blob/004ef7887a3b7fdc18277a92b114e427a70e6eeb/test/integration/test_review_app.py#L32). It is not ideal as it makes tests more complicated, and requires maintenance as the decorator's code evolves.

Another potential solution is to have the test undo the decorating. Here's [a way to obtain the original function](https://stackoverflow.com/questions/32697596/how-to-remove-the-effects-of-a-decorator-while-testing-in-python). Here's [a way to mock the decorator](https://dev.to/stack-labs/how-to-mock-a-decorator-in-python-55jc
). But it also makes the tests complicated, and it seems hacky.

That got me thinking whether the decorator is really required, and it looks not. Looks like the decorator can be replaced by a normal function.

This PR proposes to convert `command.py#comand()` into a normal wrapper function. CLI layer calls the wrapper function directly and supplies the Hokusai command function as the first argument. Much of the logic in `command.py#comand()` is retained. It seems that this is essentially what the decorator does. The difference is just that the decorator creates a new function behind the scenes, whereas now there's an extra function (the wrapper) in the call chain.

This PR also proposes to always print stack trace for any exception that is not a [HokusaiError](https://github.com/artsy/hokusai/blob/004ef7887a3b7fdc18277a92b114e427a70e6eeb/hokusai/lib/command.py#L26C14-L26C26), that is, unanticipated exceptions. Our experience is that stacktrace is always required for troubleshooting such an exception.

The changes affect almost all Hokusai commands, but the human interface has not changed, with the exception that [always_verbose](https://github.com/artsy/hokusai/blob/004ef7887a3b7fdc18277a92b114e427a70e6eeb/docs/hokusai_files.md?plain=1#L71) and [DEBUG=1](https://github.com/artsy/hokusai/blob/004ef7887a3b7fdc18277a92b114e427a70e6eeb/README.md?plain=1#L225) no longer influences exception handling.

I have spot checked a bunch of commands and they all worked. In the future we should ensure every command has a smoke test, so that such changes can be comprehensively tested.


test/integration/test_kubernetes.py
```diff
@@ -81,45 +75,35 @@ def test_03_k8s_status(self, mocked_sys_exit):
                                 content_type="application/x-amz-json-1.1")
         with captured_output() as (out, err):
             kubernetes.k8s_status(TEST_KUBE_CONTEXT, True, False, False, False, filename=self.__class__.kubernetes_yml)
-            mocked_sys_exit.assert_called_once_with(0)
             self.assertIn('Resources', out.getvalue().strip())
             # self.assertIn('deployment.apps/hello-web', out.getvalue().strip())
             # self.assertIn('service/hello-web', out.getvalue().strip())
```
cavvia, line 80: Why is there all this commented out code in our tests?

test/unit/test_command_wrapper.py
```diff
@@ -3,24 +3,23 @@
 from test import HokusaiUnitTestCase
 from test.utils import captured_output
 
-from hokusai.lib.command import command
+from hokusai.lib.command_wrapper import wrap
```
cavvia, line 6: I'm not sure why we need to change the package and the semantics of this function... `wrap` seems too generic to me. Is there something wrong with just keeping it as `command`? Either that or `execute` would seem to me more meaningful than `wrap`.

cavvia (commented):
Makes sense and the changes look good to me but I think the semantics of `wrap` needs improving as it's too generic. I don't see why we can't just keep `command` as the name of the function (can also keep the package path as well). That would appear to have a more meaningful semantics.

test/unit/test_command_wrapper.py
artsyjian: yeah let's stick with `command`. updated.

test/integration/test_kubernetes.py
artsyjian: a [5 year old commit](https://github.com/artsy/hokusai/commit/33b1d6f3f4597035e3c6d1d65dd793d5ee086d68). sounds like we wanted these assertions but they caused some stdout/err problems.

artsyjian: i am actually also working on creating a comprehensive set of integration tests. so i'll likely run into that problem and i'll address it then.

cavvia (approved):
Looks good to me now!

artsyjian:
this looks good, merging!

Issue #399, opened 2024-01-19
tests do not catch CLI errors


artsyjian:
Integration tests do not exercise the CLI, thus missing errors ([example](https://github.com/artsy/hokusai/pull/394)).

Would like to start having CLI level tests.

PR #400, opened 2024-01-19
feat: tests exercise cli

artsyjian:
to resolve https://github.com/artsy/hokusai/issues/399

major changes to integration tests:

- from now on, tests exercise hokusai cli.
- use two minikube instances, mirroring our staging/prod.
- introduce [hokusai-integration-test github repo](https://github.com/artsy/hokusai-integration-test). tests operate on hokusai files in that repo, instead of the fixtures in hokusai repo.
- introduce integration_test_local.sh script which automates local test-running, along with 'make integration-local'.
- add tests for the following hokusai cli commands:
  - version
  - check
  - staging create/status/update/refresh/delete
  - review_app setup/create/list
- delete the following tests as they are now superceded:
  - test/smoke/test_review_app.py
  - old test/integration/test_review_app.py
  - test/integration/test_kubernetes.py
- move test/integration/test_ecr.py's tests to test/unit/test_services/test_ecr.py.
- attach [hokusai-project circleci context](https://app.circleci.com/settings/organization/github/artsy/contexts/f326ddd3-84c0-4254-b446-cd66abd526fd?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgithub%2Fartsy) to test_integration build step. the context contains aws keys of [hokusai-integration-test iam user](https://us-east-1.console.aws.amazon.com/iam/home?region=us-east-1#/users/details/hokusai-integration-test-user?section=permissions) who has permissions required by the tests.

minor changes to integration tests:

- bump kubectl version, to match our current staging/prod version.
- bump minikube k8s version to match our next target staging/prod version.
- bump minikube version, because the current is too old. 
- use python 3.10.2 instead of 3.9. it reduces build time as circleci container includes 3.10.2.
- update templates in hokusai/templates/hokusai dir.
- add gitpython package, for tests to clone github repo.

future todo:

- check in hokusai-integration-tes iam user and permissions, in our infra repo.
- add integration test for every hokusai cli command.


Makefile
```diff
@@ -15,14 +15,16 @@ dependencies:
 	# pin version due to https://github.com/python-poetry/poetry/issues/7184
 	pip install poetry==1.2.2 --quiet --ignore-installed
 	poetry --version
-	poetry install --no-root
+	poetry install
```
artsyjian, line 18: turns out --no-root prevents installing Hokusai itself.

.circleci/config.yml
```diff
@@ -42,41 +42,70 @@ jobs:
...
-      K8S_VERSION: v1.19.16
+      KUBECTL_VERSION: v1.20.15
+      K8S_VERSION: v1.21.14
       KUBECONFIG: /home/circleci/.kube/config
-      MINIKUBE_VERSION: v1.22.0
+      MINIKUBE_VERSION: v1.31.2
       MINIKUBE_WANTUPDATENOTIFICATION: false
       MINIKUBE_WANTREPORTERRORPROMPT: false
       MINIKUBE_HOME: /home/circleci
-      CHANGE_MINIKUBE_NONE_USER: true
```
artsyjian, line 51: not required anymore now that minikube is not run with sudo

Makefile
cavvia: right... it's a slightly strange name for that option. I assume `make dependencies` was designed to only install the dependencies, primarily for the CI/testing use case. We generally build the project package when running tests, not sure if your tests assume that. It's relevant since some module paths may differ based on the assumption.

Makefile
```diff
@@ -15,14 +15,16 @@ dependencies:
...
 
 test:
 	coverage run -m pytest test/unit
-	coverage run --omit="test/*" -m unittest discover test.smoke
 
 integration:
-	coverage run --omit="test/*" -m unittest discover test.integration
+	coverage run -m pytest test/integration
```
cavvia, line 24: Ah interesting that we're running pytest via coverage... we should look into that in our other python repos, where we run coverage separately.

test/integration/conftest.py
```diff
@@ -0,0 +1,58 @@
...
+  if os.path.isdir(TEST_GIT_REPO_NAME) and os.environ.get('FORCE_CLONE') != '1':
+    return
+  shutil.rmtree(TEST_GIT_REPO_NAME, ignore_errors=True)
+  git.Git(".").clone(f"https://github.com/artsy/{TEST_GIT_REPO_NAME}.git")
+
+
+## autouse fixtures
+
+@pytest.fixture(scope="session", autouse=True)
+def cd_into_test_git_repo():
```
cavvia, line 53: Good use of `autouse`! It's interesting that for pytest the term `fixture` denotes any setup step, whereas in other frameworks, like rspec, the term has a narrower definition, restricted just to "test data", and setup and teardown are handled separately. I think your use of fixtures is in-line with pytest convention.

test/integration/conftest.py
```diff
@@ -0,0 +1,58 @@
...
+  os.chdir(TEST_GIT_REPO_NAME)
+
+@pytest.fixture(autouse=True)
+def test_git_repo_name():
```
cavvia, line 57: Not convention to have a fixture named like a test... this doesn't seem to assert anything so probably shouldn't be a `test_` function.

test/integration/test_staging.py
```diff
@@ -0,0 +1,97 @@
...
+    if resp.returncode != 0:
+      print(resp.stderr)
+    assert resp.returncode == 0
+    assert 'deployment.apps/hokusai-integration-test-web unchanged' in resp.stdout
+    assert 'service/hokusai-integration-test-web unchanged' in resp.stdout
+    assert 'Updated Kubernetes environment' in resp.stdout
+
+def describe_refresh():
+  def it_times_out():
+    # expect timeout due to minikube lacking ECR image pull permission
```
cavvia, line 66: This is a little unorthodox... how long does this test take to run?

cavvia (commented):
@artsyjian I'm struggling to understand why the config files for the minikube test environment need to live in a separate repo... can you clarify the reasoning for that?

test/integration/conftest.py
```diff
@@ -0,0 +1,58 @@
...
+  # 127.0.0.1 = local
+  # 192.168.x.y = circleci
+  if not (
+    control_plane_address.startswith('127.0.0.1') or
+    control_plane_address.startswith('192.168')
+  ):
+    msg = f'Error: {context} context possibly points to a real cluster. Is it a Minikube instance?'
+    pytest.exit(msg)
+
+def pytest_configure(config):
```
cavvia, line 36: This file performs setup for integration tests... perhaps a more descriptive file name than `conftest` would be helpful? Something like `test/integration/setup.py` for example? If you need pytest to auto-discover it you can tweak config for that perhaps.

Makefile
artsyjian: ah, thanks for pointing that out. since the step is named `dependencies`, `--no-root` seems appropriate. i should perhaps rename it, or create a separate `make` step for the integration tests.

test/integration/conftest.py
artsyjian: yeah, this is an unusual fixture, it `cd`'s into dir before every test, not global.

test/integration/conftest.py
artsyjian: good catch. shall rename. it's mean to convey 'this fixture returns the name of the git repo that is used for the tests...'

test/integration/test_staging.py
artsyjian: yeah, initially i wanted the test to confirm that the deployment is successfully refreshed, but ran into trouble with ECR authentication. hokusai running in circleci container has access to circleci context with aws keys, so hokusai itself can authenticate to ECR. but those creds are not accessible by k8s running inside minikube, and k8s needs them for image pull. so we can configure minikube/k8s with the creds, not sure if that's good idea in regards to security, but i came to the conclusion that for the purpose of testing 'hokusai staging refresh' it seems good enough that the test sees `Refreshing...`, so letting the test time out (the timeout is b/c of minikube/k8s retrying/failing image pull)

artsyjian:
> @artsyjian I'm struggling to understand why the config files for the minikube test environment need to live in a separate repo... can you clarify the reasoning for that?

great question! ultimately i want to be able to test hokusai's [git tag push steps](https://github.com/artsy/hokusai/blob/6a8f793f5b3728e43de58d28448524ff6b53cc12/hokusai/services/deployment.py#L191), so will have to have a real github repo for the test project, i think.

test/integration/conftest.py
artsyjian: agreed. i am under the impression that the file has to be named `conftest.py`.

test/integration/test_staging.py
artsyjian: on second thought, seems best for test to confirm successful refresh.

test/integration/conftest.py
cavvia: Ah I see the [doc](https://docs.pytest.org/en/7.1.x/reference/fixtures.html#:~:text=The%20conftest.py%20file%20serves,pytest%20will%20automatically%20discover%20them) now, I haven't used one of those files before. We usually just load fixtures in from wherever we choose to place them. Ignore my comment in that case.

test/integration/test_staging.py
cavvia: If the creds are already in the circle ci context what is the security issue around propagating them to minikube? My assumption is that it's local to the circle container as well, in which case I agree just running it with the creds and testing successful refresh would probably be preferable to waiting for a timeout. Alternatively, setting a very low timeout for the test and just checking that the command is issued (rather than testing the timeout itself) is acceptable I think.

cavvia:
That makes sense, thanks for the clarification @artsyjian. Do you want to update the ecr pull test? Otherwise looks good to me and am happy to approve.

cavvia (approved):
Given that the time out test is executing very quickly I'm happy to merge this in... you can follow up if you'd like to clean up that one integration test by providing credentials to complete a test ecr pull.

artsyjian:
thanks @cavvia. yeah i have yet to address the 3 outstanding comments. i will follow up with another pr.

artsyjian:
ah, thanks for fixing poetry.lock merge conflict.

PR #401, opened 2024-01-22
feat: use tmp_path fixture in pytest tests

artsyjian:
Replace this clunky construct:

```
with tempfile.TemporaryDirectory() as tmpdir
```

with pytest's [tmp_path fixture](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html).


amonkhouse (approved):
hadn't seen this fixture before, looks good!

PR #409, opened 2024-05-22
feat: bump python and xcode versions

artsyjian:
Builds are suddenly failing.

- [test_docker_build](https://app.circleci.com/pipelines/github/artsy/hokusai/1264/workflows/1095bdf0-5ba5-45e9-9ff2-0ef0f6ead2f7/jobs/5432?invite=true#step-105-12_76) fails with Segmentation Fault.
- [test_macos](https://app.circleci.com/pipelines/github/artsy/hokusai/1266/workflows/6fcbffd4-f1f7-40e9-90b1-36b1bae50628/jobs/5440?invite=true#step-105-926_39) fails at installing backcall.

The builds currently use Python 3.9. Bumping Python to 3.10 makes the errors go away. Let's do that.

The MacOS steps currently use Xcode 13.4.1 which [maps to MacOS 12.3.1](https://circleci.com/docs/using-macos/#supported-xcode-versions-intel) which is quite old. Let's bump Xcode to 15.3.0, the latest supported by CircleCI for Intel chip.

Also bump `kubectl` version to 1.22, to match our Kubernetes clusters.


.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
   test_macos:
     macos:
-      xcode: "13.4.1"
+      xcode: "15.3.0"
```
brainbicycle, line 16: macos intel resource were EOL'd I believe so I think you are going to get a sillicon resource regardless: https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718

if that is true might want to just bump to lates 15.4.0

brainbicycle (approved):
minor comment about xcode version to choose, lgtm!

.circleci/config.yml
artsyjian: sounds like we have to manually define a [resource class key](https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718#how-to-migrate-to-m1-4) to use apple silicon.

will hokusai binary built for apple silicon work on intel? i am still on intel, as probably are some other developers. we might have to somehow do 2 builds.

brainbicycle: hmm, I am confused, reading that issue it sounds to me like you already cannot use non sillicon mac os executors but it seems like that is not the case and it is using the old medium gen 2 images: https://app.circleci.com/pipelines/github/artsy/hokusai/1275/workflows/6af079ea-16a5-4645-be09-a0a435f7b5d5/jobs/5483
I suppose we can just continue using this but it might start to fail at some point in the future.

brainbicycle: Oh June 28, 2024 I can't read. I will put it on the disaster calendar

artsyjian:
changed assignee to @brainbicycle b/c @joeyAghion is ooo.

artsyjian:
Merging into `main` creates [Beta](https://github.com/artsy/hokusai?tab=readme-ov-file#beta-release) (binaries and Docker image). [Several Artsy projects](https://github.com/search?q=org%3Aartsy+%22hokusai%2Fbeta%22&type=code) are configure to use Beta Docker image for their builds. If they continue to work, we will create a [canonical release](https://github.com/search?q=org%3Aartsy+%22hokusai%2Fbeta%22&type=code).

Issue #408, opened 2024-05-13
Review App creation fails due to embedded Jinja


artsyjian:
While [embedded Jinja](https://github.com/artsy/hokusai-sandbox/blob/a22c1fd6ab57c87653bca59cd83ea5da21ddd1da/hokusai/staging.yml.j2#L18C9-L18C18) can be escaped using `{% raw %}`, it still causes problem for Review App creation.

The problem happens in 2 steps:

- Review app setup

  ```
  artsy:hokusai-sandbox jxu$ /Users/jxu/.pyenv/versions/hokusai/bin/hokusai review_app setup foo
  Created hokusai/foo.yml
  ```

  In this step, Hokusai [loads staging.yml spec](https://github.com/artsy/hokusai/blob/c480867b842aadfb2a9604aada4a7fab71922017/hokusai/commands/namespace.py#L18) and renders any Jinja templates in the spec. This removes the raw delimiters, so that ES resource’s embedded Jinja template is no longer escaped. The produced Review App spec `foo.yml` contains:

  ```
      template:
        engineVersion: v2
        templateFrom:
        - literal: '{{ range $key, $value := . }}
  
            {{$key}}: {{$value | fromJson | values | first}}
  
            {{ end }}
  
            '
          target: Data
  ```

- Review app create

  ```
  artsy:hokusai-sandbox jxu$ /Users/jxu/.pyenv/versions/hokusai/bin/hokusai review_app create foo
  ERROR: Rendering template raised error <class 'jinja2.exceptions.TemplateSyntaxError'> <message 'unexpected char '$' at 682'>
  ```

  The error occurs because Hokusai attempts to Jinja render `foo.yml` ES resource.

PR #410, opened 2024-05-24
feat: prevent review app specs from being jinja rendered

artsyjian:
Solves https://github.com/artsy/hokusai/issues/408

A review app spec does not have to be Jinja rendered because any Jinja template was already rendered when the spec was created during `hokusai review_app setup`. During spec creation, Hokusai copied Staging spec and rendered Jinja templates.

When Staging spec has Jinja template escaped with `{% raw %}`, the double rendering causes  `hokusai review_app create` to break.

This PR proposes to eliminate the second rendering. To achieve that, a `render_template` flag is introduced and passed down the call chain. We default it to `True` so there's no impact on non-review-app commands (e.g. `hokusai staging ...`, `hokusai production ...`). Review app commands set it to `False`, so that the review app spec is not rendered.

This PR also replaces legacy `test/unit/test_yaml_spec.py` with a Pytest-format version that is more comprehensive.

### DIFF
--- a/test/unit/test_services/test_yaml_spec.py
+++ b/test/unit/test_services/test_yaml_spec.py
@@ -66,6 +66,10 @@ def describe_yaml_spec():
       mocker.patch('hokusai.services.yaml_spec.NamedTemporaryFile', return_value=f)
       file_name = obj.to_file()
       assert file_name == f.name
+      content = ''
+      with open(file_name, 'r') as f:
+        content = f.read().strip()
+      assert content == 'foo'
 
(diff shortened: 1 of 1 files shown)


joeyAghion (approved):
If I understand correctly, this means review apps' specs will be "rendered" once at create time, while standard deployments will be rendered each time they're needed. Do we need to be careful about changing template inputs, since review apps won't get the updates? (In my memory, we only change those inputs very, very rarely.)

PR #417, opened 2024-06-26
Switch CI MacOS builds to M1 and continue to build x86_64 binary

artsyjian:
Currently, CI MacOS builds run on Intel MacOS and they output x86_64 binaries which work on Intel MacOS as well as on Apple silicon MacOS that has [Rosetta](https://support.apple.com/en-us/102527) installed.

CircleCI is [deprecating Intel MacOS](https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718). Going forward, the builds have to run on Apple silicon.

A binary built on Apple silicon does not work on Intel. When running an Apple-silicon-built binary on Intel, this error occurs:
```
artsy:hokusai-m1 jxu$ ./hokusai
-bash: ./hokusai: Bad CPU type in executable
```

In this PR, we do two things. First, we switch the builds to Apple silicon (M1), in anticipation of CircleCI's deprecation. Second, we make the builds output x86_64 binaries, so that we continue to have a single binary that works on both platforms.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,8 @@ jobs:
       - run: pyenv rehash
       - run: make test
       - run: coverage report
-      - run: make pyinstaller-build-onedir ARTIFACT_LABEL=beta
+      - run: make pyinstaller-build-onedir ARTIFACT_LABEL=test
+      - run: make publish-to-s3 ARTIFACT_LABEL=test
 
   test_integration:
     machine:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -420,6 +420,8 @@ workflows:
             branches:
               ignore: release
       - test_macos:
+          context:
+          - hokusai
           filters:
             branches:
               ignore: release
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,7 @@ jobs:
           name: install awscli
           command: |
             curl "https://awscli.amazonaws.com/AWSCLIV2.pkg" -o "AWSCLIV2.pkg"
+            sudo softwareupdate --install-rosetta --agree-to-license
             sudo installer -pkg AWSCLIV2.pkg -target /
       - run: make hokusai
       - run: pyenv rehash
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -14,29 +14,17 @@ jobs:
...
-            pyenv install 3.10
-            pyenv local 3.10
-            pip install --upgrade pip
-      - run:
-          name: install awscli
-          command: |
-            curl "https://awscli.amazonaws.com/AWSCLIV2.pkg" -o "AWSCLIV2.pkg"
-            sudo installer -pkg AWSCLIV2.pkg -target /
-      - run: make hokusai
-      - run: pyenv rehash
```
artsyjian, line 36: the above steps are repeated 3x for each MacOS job (test, release-beta, release), so let's move them to a script.

.circleci/config.yml
```diff
@@ -14,29 +14,17 @@ jobs:
...
+            ./scripts/ci_macos_build_hokusai.sh
       - run: make test
       - run: coverage report
-      - run: make pyinstaller-build-onedir ARTIFACT_LABEL=beta
+      - run: make pyinstaller-build-onedir ARTIFACT_LABEL=test
+      - run: make publish-to-s3 ARTIFACT_LABEL=test
```
artsyjian, line 27: upload binary to S3 so we can download and test on our locals.

.circleci/config.yml
```diff
@@ -413,6 +375,8 @@ workflows:
             branches:
               ignore: release
       - test_macos:
+          context:
+          - hokusai
```
artsyjian, line 379: provides permissions to S3.

Makefile
```diff
@@ -40,6 +40,7 @@ pyinstaller-build-onefile: # for linux
 # for mac (because build-onefile's binary runs too slow on mac)
 pyinstaller-build-onedir:
 	pyinstaller \
+	  --target-architecture=x86_64 \
```
artsyjian, line 43: build x86_64 binary.

scripts/ci_macos_build_hokusai.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+# build Hokusai in CI environment
+# for use on Apple silicon MacOS
+
+set -xeo pipefail
+
+# install rosetta
+time sudo softwareupdate --install-rosetta --agree-to-license
```
artsyjian, line 9: allows the subsequent steps to be done in x86_64.

.circleci/config.yml
joeyAghion: If they're only repeated _in this file_, you might be able to just use a YAML alias instead of a new file.

joeyAghion (approved):

PR #419, opened 2024-08-15
let "hokusai staging/production/review_app run" commands derive overrides from deployment spec

artsyjian:
[PHIRE-1103]

The Kubernetes-related `run` commands (i.e. staging/production/review_app run) execute `kubectl run...` under the cover. [kubectl run](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_run/#synopsis) has an `--overrides` option which [Hokusai uses](https://github.com/artsy/hokusai/blob/7c811f9a26254f3fbdad898ca3eb71346165048a/hokusai/services/command_runner.py#L110) to feed in a bunch of customizations. While most of the customizations are based on command line options specified by user, some are not (e.g. [envFrom](https://github.com/artsy/hokusai/blob/7c811f9a26254f3fbdad898ca3eb71346165048a/hokusai/services/command_runner.py#L59C10-L59C17)) and they are hardcoded in Hokusai code. The hardcoded customizations can actually be taken from [Deployment specs](https://github.com/artsy/hokusai-sandbox/blob/7df6807bc2ff75f441ea989941ca49b2d0847d7b/hokusai/staging.yml#L42), but Hokusai currently does not parse them.

We plan to add [initContainers](https://github.com/artsy/hokusai-sandbox/blob/7df6807bc2ff75f441ea989941ca49b2d0847d7b/hokusai/staging.yml#L33) to Artsy projects' specs. The initContainers should be included in `run` commands. If we were to continue the current pattern, we would hardcode initContainers in Hokusai as well, but that is not sustainable. For one, the presence of an initContainer or its spec is quite org-specific. Second, in the future we may add even more resources to projects' pod specs that should be included by `run` commands.

Therefore, this PR proposes a new approach to doing the overrides. Since almost all Artsy projects have at least one Kubernetes [`Deployment` resource](https://github.com/artsy/hokusai-sandbox/blob/7df6807bc2ff75f441ea989941ca49b2d0847d7b/hokusai/staging.yml#L3), let's base the overrides on that. We extract the [pod spec](https://github.com/artsy/hokusai-sandbox/blob/7df6807bc2ff75f441ea989941ca49b2d0847d7b/hokusai/staging.yml#L32) from `Deployment` spec, then clean it up, removing fields (e.g. readinessProbe) that are meaningless to `run`. Then we apply user's command line options on top of that.

We introduce a new `hokusai/config.yml` config called `run-template` that allows a user to specify the name of the Deployment. If this config is not present, Hokusai attempts to use the first Deployment found in the yamls.

All changes in this PR should be backward compatible. Hokusai projects that don't have the new config should see no changes.

[PHIRE-1103]: https://artsyproduct.atlassian.net/browse/PHIRE-1103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/hokusai_files.md
```diff
@@ -56,19 +56,21 @@ Please see [hokusai configure command](./Command_Reference.md#configuring-hokusa
...
 - `run-constraints`: <list of kubernetes label selector strings> - Constrain run containers to Kubernetes nodes matching the label selectors in the form `key=value` by setting the `nodeSelector` field on the container's spec. Bound to the `--constrint` option for `hokusai [staging|production|review_app] run` as well as containers run via the `--migration` flag as well as `pre-deploy` / `post-deploy` hooks triggered by `hokusai [staging|production|review_app] deploy` or `hokusai pipeline promote`.  Falls back to the `HOKUSAI_RUN_CONSTRAINTS` env var, in which case a list is parsed from a comma-delimited string.
-- `always-verbose`: <boolean> (optional) - Always print verbose output.  Bound to the `--verbose` option for various commands, and falls back to the `HOKUSAI_ALWAYS_VERBOSE` env var.
+- `run-template`: <string> (optional) - `hokusai [staging|production|review_app] run` creates a pod to run the user-specified command. This config specifies name of the Deployment in `hokusai/staging.yml` or `hokusai/production.yml` to model the run-pod's spec after. If not specified, Hokusai uses the first Deployment found in those files.
```
artsyjian, line 69: new

docs/hokusai_files.md
```diff
@@ -56,19 +56,21 @@ Please see [hokusai configure command](./Command_Reference.md#configuring-hokusa
...
+- `run-tty`: <boolean> (optional) - Attach the terminal to your shell session when invoking `hokusai [staging|production|review_app] run`.  Bound to the `--tty` option for this command, and falls back to by the `HOKUSAI_RUN_TTY` env var.
+- `secrets-file`: <string> (optional): A project may rely on secret app configs that are loaded from a file into the shell environment of its app container. This config specifies the path to that file. The content of the file must be in `export KEY='VALUE'` form. `hokusai [staging|production|review_app] run` reads this config and takes care of `source`ing the file, in addition to running the user-specified command.
```
artsyjian, line 71: new

hokusai/lib/common.py
```diff
@@ -70,6 +95,28 @@ def k8s_uuid():
...
+
+  return:
+  {
+    'key1': 'value1',
+    'key2': 'value2',
+    ...
+  }
+  '''
+  return {
+    key_value.split('=', 1)[0]: key_value.split('=', 1)[1]
```
artsyjian, line 115: anyway to split just once?

docs/hokusai_files.md
artsyjian: Oh yes, secrets file path will likely be the same from project to project, so does seems it should be a global config. If so, we want a way to migrate projects gracefully, so that `run` still works with projects not yet using a secrets file.

mc-jones:
>If this config is not present, Hokusai attempts to use the first Deployment found in the yamls.

Wondering if this could present weird edge cases? I don't think we have a standard order for deployments, so is there a potential scenario where the first deployment's image wouldn't support the run command being passed?

artsyjian:
> Wondering if this could present weird edge cases? I don't think we have a standard order for deployments, so is there a potential scenario where the first deployment's image wouldn't support the run command being passed?

For sure I think. Agreed that this approach is not very reliable. Just that it seems the sainest fallback when `run-template` is not specified. Surveyed my local repos and among all the Kubernetes apps, only these don't have a `-web` deployment:
```
frequency
fulcrum
opstools
```

So for our environment, we can set `run-template: <project>-web` for almost all projects.

artsyjian:
Discussed with @joeyAghion, we may have better ideas for injecting `secrets-file`, so I will remove that part from this PR and scope down the PR to just  [PHIRE-1103](https://artsyproduct.atlassian.net/browse/PHIRE-1103) - changing overrides approach.

docs/hokusai_files.md
artsyjian: removed `/secrets/secrets` related code. will follow up on future PR.

artsyjian:
Looks like this should be good to merge?

mc-jones (approved):

PR #425, opened 2024-09-06
feat: review app create/delete namespace outside of yaml

artsyjian:
Solves [PHIRE-1160]

Currently, the Kubernetes namespace for a review app is [spec'd in the review app's YAML](https://github.com/artsy/hokusai/blob/22c8ce02850b38497fa6842818b1b154035201a5/hokusai/commands/namespace.py#L27), therefore, created at Kubectl create/apply on the YAML.

This presents a problem with creating review apps for a project that uses a [project-specific Kubernetes service account](https://github.com/artsy/hokusai-sandbox/blob/d8bf76f6ee193cc78cd0f4cd61144af8c50b196a/hokusai/staging.yml#L86). For such a project, Hokusai should replicate the SA from `default` namespace to review app namespace. This should be done 1) after the namespace is created, and 2) before review app YAML is applied. That is impossible when namespace is part of YAML.

This PR breaks namespace spec out of the YAML - create namespace in [review_app setup](https://github.com/artsy/hokusai/blob/22c8ce02850b38497fa6842818b1b154035201a5/hokusai/cli/review_app.py#L22), not [review_app create](https://github.com/artsy/hokusai/blob/22c8ce02850b38497fa6842818b1b154035201a5/hokusai/cli/review_app.py#L35).

This PR includes a lot of refactoring and shoring up of tests.

This PR should not introduce any user-facing changes. It's just preparation for adding service account replication logic which will be done in a follow up PR.

[PHIRE-1160]: https://artsyproduct.atlassian.net/browse/PHIRE-1160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hokusai/cli/review_app.py
```diff
@@ -18,12 +18,18 @@ def review_app(context_settings=CONTEXT_SETTINGS):
...
+  default='hokusai/staging.yml',
+  help="The source yaml file from which to create the new resource file (default: hokusai/staging.yml)"
+)
 def setup(app_name, verbose, source_file):
-  """Setup a new review-app - create a Yaml file based on APP_NAME and --source-file"""
+  """Setup a new Review App"""
   set_verbosity(verbose)
   command(
-    hokusai.create_new_app_yaml,
+    hokusai.setup_review_app,
```
artsyjian, line 32: since `setup` now does more than just creating yaml - it creates namespace as well.

hokusai/cli/review_app.py
```diff
@@ -51,11 +57,10 @@ def delete(app_name, verbose):
   """Deletes the Kubernetes based resources defined in ./hokusai/{APP_NAME}.yml"""
   set_verbosity(verbose)
   command(
-    hokusai.k8s_delete,
+    hokusai.delete_review_app,
```
artsyjian, line 60: deserves its own review app command, since now it does namespace deletion outside of k8s_delete.

hokusai/cli/review_app.py
```diff
@@ -178,7 +209,7 @@ def deploy(app_name, tag, migration, constraint, git_remote, timeout, update_con
 def list(verbose):
   """List existing review apps of the project"""
   set_verbosity(verbose)
-  hokusai.list_namespaces(KUBE_CONTEXT, labels=f'app-name={config.project_name},app-phase=review')
+  hokusai.list_review_apps(KUBE_CONTEXT, labels=f'app-name={config.project_name},app-phase=review')
```
artsyjian, line 212: best not let user know review app implementation detail - that it uses namespaces.

hokusai/lib/common.py
```diff
@@ -238,6 +251,14 @@ def smart_str(s, newline_before=False, newline_after=False):
...
   return s
 
+def unlink_file_if_not_debug(path):
+  ''' if DEBUG is off, unlink specified file '''
+  if not os.environ.get('DEBUG'):
+    try:
+      os.unlink(path)
+    except:
+      pass
+
```
artsyjian, line 261: a lib for this logic instead of it sprinkled in many places

hokusai/lib/common.py
```diff
@@ -294,7 +317,21 @@ def validate_key_value(key_value):
...
     else:
       print_yellow("==> hokusai exec `%s`" % msg, newline_after=True)
   return msg
+
+def write_temp_file(data, dir):
+  ''' write data to temp file, return file path '''
+  obj = NamedTemporaryFile(delete=False, dir=dir, mode='w')
+  obj.write(data)
+  obj.close()
+  return obj.name
```
artsyjian, line 333: until now, this logic is sprinkled in many places.

hokusai/services/configmap.py
```diff
@@ -1,72 +1,77 @@
...
       'kind': 'ConfigMap',
       'metadata': self.metadata,
       'data': {}
     }
 
-  def _to_file(self):
-    f = NamedTemporaryFile(delete=False, dir=HOKUSAI_TMP_DIR, mode='w')
-    f.write(json.dumps(self.struct))
-    f.close()
-    return f
```
artsyjian, line 35: not configmap specific, moved to lib.

mc-jones (approved):
Looks great, with my default disclaimer, that someone with better `python` knowledge will be able to comment more insightfully on the implementation. 

My only question is a somewhat philosophical one... By reworking the command to focus on review apps, are we continuing to move `hokusai` towards a more `artsy` specific implementation, rather than a generalized wrapper around `k8s`?

artsyjian:
> My only question is a somewhat philosophical one... By reworking the command to focus on review apps, are we continuing to move `hokusai` towards a more `artsy` specific implementation, rather than a generalized wrapper around `k8s`?

That's always something to be vigilant about.

The previous PRs seem okay in that regard. [This one](https://github.com/artsy/hokusai/pull/419/files#diff-65944c1c179b582e64c48923af70c36689b944e535ceb7496f3345acf3bb25beL62) actually made Hokusai less Artsy-specific, as we stopped hardcoding our ConfigMap naming convention. 

This PR moves Namespace creation out of Review App YAML, which doesn't seem to make Hokusai more or less Artsy-specific. Having Namespace spec in the YAML seems a flaw, as it [forces](https://github.com/artsy/hokusai/blob/main/docs/Review_Apps.md) us to replicate ConfigMap (by `hokusai review_app env copy <name>`) _after_ the YAML is applied (by `hokusai review_app create <name>`). ConfigMap should be created first because it's referred to in the YAML, but it can't be created first because the namespace does not exist (it gets created when the YAML is applied). As it is now, the pods spawned by `hokusai review_app create <name>` go into an error state, then when ConfigMap is replicated, they recover. That's really a bug.

In the next PR I plan to have Hokusai parse the YAML for any references to ConfigMap and ServiceAccount. Those are really pre-requisite resources, so it seems fine for Hokusai to just replicate them. `hokusai review_app env copy <name>` should no longer be required. This approach also doesn't seem to make Hokusai more or less Artsy-specific. If an org's YAML doesn't refer to ConfigMap or ServiceAccount, Hokusai just doesn't replicate.

mc-jones:
> That's always something to be vigilant about.

Makes sense. Appreciate the details run through. 🙏

joeyAghion (approved):

PR #426, opened 2024-09-19
feat: review_app setup command copies configmaps and service accounts

artsyjian:
[PHIRE-1153]

[hokusai review_app env copy](https://github.com/artsy/hokusai/blob/main/docs/Review_Apps.md) command allows user to copy a project's Staging configmap from default namespace to review app namespace. Since review app Yaml is [derived from staging Yaml](https://github.com/artsy/hokusai/blob/2fba5a5ebaa38e43ff1b7727d9616b740e96e7a0/hokusai/commands/review_app.py#L23), any configmap references are carried over. Referenced configmaps must be copied.

That's for configmaps. Now we we have something similar for serviceaccounts. We are introducing [project-specific serviceaccounts](https://github.com/artsy/hokusai-sandbox/blob/d8bf76f6ee193cc78cd0f4cd61144af8c50b196a/hokusai/staging.yml#L86) for Artsy projects. Like configmaps, serviceaccounts must also be copied.

We can introduce, say, a `serviceaccount copy` command, but that seems tedious.

This PR proposes a new approach. If any referenced configmaps or serviceaccounts do not exist in review app namespace, pods there cannot run. Rather than waiting for user to run copy commands, let hokusai just copy automatically, as part of [review_app setup](https://github.com/artsy/hokusai/blob/2fba5a5ebaa38e43ff1b7727d9616b740e96e7a0/hokusai/cli/review_app.py#L28). Let Hokusai parse review app Yaml to find out what configmaps/serviceaccounts to copy.

This PR deprecates `env copy` command. The command now merely prints a warning telling user to stop using it, allowing time for any [user automation that calls the command](https://github.com/search?q=org%3Aartsy%20%22hokusai%20review_app%20env%20copy%22&type=code) to be updated. A future PR will actually remove the command.

[PHIRE-1153]: https://artsyproduct.atlassian.net/browse/PHIRE-1153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hokusai/cli/review_app.py
```diff
@@ -248,31 +248,6 @@ def env(context_settings=CONTEXT_SETTINGS):
...
-  """Copies the app's environment config map to the namespace {APP_NAME}"""
-  set_verbosity(verbose)
-  command(
-    hokusai.k8s_copy_config,
-    KUBE_CONTEXT,
-    clean_string(app_name),
-    name=configmap
-  )
-
-
```
artsyjian, line 275: `hokusai review_app env copy` command

hokusai/lib/common.py
```diff
@@ -57,6 +57,45 @@ def clean_dict(d1, fields_to_keep):
...
+      'bb': 3
+    }
+  }
+  '''
+  for key in keys:
+    if key in d1:
+      if keys[key] == {}:
+        del d1[key]
+      else:
+        delete_keys(d1[key], keys[key])
```
artsyjian, line 97: recursion, yeah.

hokusai/services/configmap.py
```diff
@@ -53,7 +53,8 @@ def delete(self, key):
   def destroy(self):
     ''' delete configmap in Kubernetes '''
     shout(
-      self.kctl.command("delete configmap %s" % self.name)
+      self.kctl.command("delete configmap %s" % self.name),
+      print_output=True
```
artsyjian, line 57: that makes output from kubectl print.

pyproject.toml
```diff
@@ -40,6 +40,9 @@ pytest-watch = "^4.2.0"
 twine = "^3.8.0"
 wheel = "^0.37.1"
 
+[tool.poetry.group.dev.dependencies]
+pytest-order = "^1.3.0"
```
artsyjian, line 44: to run integration tests in order

scripts/integration_test_local.sh
```diff
@@ -62,14 +62,14 @@ function create_real_docker_config_json() {
...
+  if ls "$KUBE_CONFIG_PATH" &>/dev/null
   then
-    echo "Backing up $KUBE_CONFIG_PATH ..."
     mv "$KUBE_CONFIG_PATH" "$KUBE_CONFIG_BACKUP_PATH"
-    backed_up_kube_config='true'
+    echo "Backing up $KUBE_CONFIG_PATH..."
+    retval=0
   fi
-  echo "$backed_up_kube_config"
+  return "$retval"
```
artsyjian, line 72: been relying on `echo` to return backed_up_kube_config flag, problem is all `echo`s are returned. so switching to `return`.

test/integration/conftest.py
```diff
@@ -33,25 +48,42 @@ def exit_pytest_if_not_minikube(context):
...
-  git.Git(".").clone(f"https://github.com/artsy/{NAME_OF_TEST_GIT_REPO}.git")
+  # clone test repo if:
+  # - local repo dir does not exist, or
+  # - we force
+  if not os.path.isdir(TEST_REPO_DIR) or os.environ.get('FORCE_CLONE') == '1':
+    shutil.rmtree(TEST_REPO_DIR, ignore_errors=True)
+    git.Git(FIXTURES_DIR).clone(f"https://github.com/artsy/{NAME_OF_TEST_GIT_REPO}.git")
+  # apply service account spec
+  apply_test_repo_serviceaccount('staging', TEST_REPO_DIR)
+  apply_test_repo_serviceaccount('production', TEST_REPO_DIR)
```
artsyjian, line 69: i made [test deployment](https://github.com/artsy/hokusai-integration-test/blob/dd85543fda77594a5d69b85dfc1812c34ded9b09/hokusai/staging.yml#L40) use project-specific serviceaccount.

test/integration/conftest.py
```diff
@@ -33,25 +48,42 @@ def exit_pytest_if_not_minikube(context):
...
+  test git repo directory: f{TEST_REPO_DIR}
+  '''
+  )
 
 ## autouse fixtures
 
 @pytest.fixture(scope="session", autouse=True)
 def cd_into_test_git_repo():
-  os.chdir(NAME_OF_TEST_GIT_REPO)
+  os.chdir(TEST_REPO_DIR)
```
artsyjian, line 86: better use absolute dir

test/integration/test_staging.py
```diff
@@ -78,6 +81,7 @@ def it_reports_success():
     assert 'Waiting for deployment "hokusai-integration-test-web" rollout to finish' in resp.stdout
     assert 'deployment "hokusai-integration-test-web" successfully rolled out' in resp.stdout
 
+@pytest.mark.order(1510)
```
artsyjian, line 84: run this after review app tests. made possible by pytest-order plugin.

joeyAghion (commented):
You _could_ preserve the `env copy` command, in the interest of a more graceful upgrade. Is that too messy? It could even warn about the upcoming removal.

artsyjian:
> You _could_ preserve the `env copy` command, in the interest of a more graceful upgrade. Is that too messy? It could even warn about the upcoming removal.

makes sense. i'll do that.

pyproject.toml
joeyAghion: Is this because they're actually interdependent now?

joeyAghion (approved):

hokusai/commands/review_app.py
```diff
@@ -13,11 +13,39 @@
...
+  destination_configmap = ConfigMap(
+    'staging',
+    name=name,
+    namespace=destination_namespace
+  )
+  source_configmap.load()
+  destination_configmap.struct['data'] = source_configmap.struct['data']
+  destination_configmap.save()
+
+def copy_sa(name, destination_namespace):
```
mc-jones, line 35: Non-blocking, but perhaps its better to be more declarative here and use `copy_service_account`?

artsyjian: sure

pyproject.toml
artsyjian: yeah

hokusai/lib/common.py
mc-jones: Might just be me reading this wrong, but are we missing a base case to prevent infinite recursion here?

mc-jones (approved):
Awesome. Love simplifying the review app process. 

Just one quick question about the recursive `delete_keys` but looks good.

hokusai/lib/common.py
artsyjian: recursion should stop when `for key in keys` is exhausted at the most nested level. should be okay?

Issue #433, opened 2025-01-24
Docker Desktop for Mac removes docker-compose command


artsyjian:
Looks like starting with [Docker Desktop 4.32.0](https://github.com/docker/for-mac/issues/7345#issuecomment-2256328781), `docker-compose` command is of no more. Users will have to use `docker compose` command instead.

This breaks `hokusai dev` and `hokusai test` commands on developer locals because Hokusai shells out to `docker-compose`.

artsyjian:
also affects `hokusai build`.

PR #434, opened 2025-01-31
feat: fallback to "docker compose" when docker-compose not found

artsyjian:
To solve https://github.com/artsy/hokusai/issues/433

`docker compose` seems to be able to do everything `docker-compose` does.

This PR recommends that when `docker-compose` is not available, use `docker compose`. The following Hokusai commands use Docker Compose:
- `hokusai build`
- `hokusai dev`
- `hokusai test`

This PR includes the following refactors:

- Add verbosity option to `hokusai check` command.
- The responsibility of discovering `docker-compose` command is moved from `hokusai check` command to `Docker` class.
- All invocations of `docker-compose` are changed to invoke whichever command `Docker` class decides.
- The following spaghetti is moved into helpers:
  ```
    if filename is None:
      yaml_template = TemplateSelector().get(os.path.join(CWD, HOKUSAI_CONFIG_DIR, DEVELOPMENT_YML_FILE))
    else:
      yaml_template = TemplateSelector().get(filename)

    docker_compose_yml = YamlSpec(yaml_template).to_file()
  ```
- Minor formatting improvements.

### DIFF
--- a/hokusai/commands/development.py
+++ b/hokusai/commands/development.py
@@ -1,14 +1,14 @@
 import os
 import signal
 
-from hokusai.lib.config import HOKUSAI_CONFIG_DIR, config
+from hokusai.lib.config import HOKUSAI_CONFIG_DIR, DEVELOPMENT_YML_FILE, config
 from hokusai.lib.common import print_green, shout, EXIT_SIGNALS
 from hokusai.services.docker import Docker
 from hokusai.lib.docker_compose_helpers import generate_compose_command, get_yaml_template
 
 
 def dev_start(build, detach, filename):
-  compose_command = generate_compose_command(filename)
+  compose_command = generate_compose_command(filename, default_yaml_file=DEVELOPMENT_YML_FILE)
   def cleanup(*args):
     shout(
       f'{compose_command} -p hokusai stop',
@@ -18,7 +18,7 @@ def dev_start(build, detach, filename):
     signal.signal(sig, cleanup)
   opts = ''
   if build:
-    yaml_template = get_yaml_template(filename)
+    yaml_template = get_yaml_template(filename, default_yaml_file=DEVELOPMENT_YML_FILE)
     Docker().build(filename=yaml_template)
   if detach:
     opts += ' -d'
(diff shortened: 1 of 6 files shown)


hokusai/lib/docker_compose_helpers.py
```diff
@@ -22,3 +22,42 @@ def follow_extends(docker_compose_yml):
       extended_template = TemplateSelector().get(extended_template_path)
       rendered_templates.append(YamlSpec(extended_template).to_file())
     return rendered_templates
+
+def generate_compose_command(filename, default_yaml_file):
+  ''' return Docker Compose command '''
+  # this import when done globally causes circular import error
+  from hokusai.services.docker import Docker
```
amonkhouse, line 29: This seems like a bad pattern and is a bit confusing to follow, I think we should avoid doing this. I feel like there's two best options to change it to:

1. keep this function here, but alter it to take an extra arg, something like `compose_command`, and use that in place of `Docker.compose_command()`.
2. move `compose_command` from docker.py into this helpers file. I personally feel like it would fit well in this file but you know this code better than me!

hokusai/services/docker.py
```diff
@@ -1,32 +1,39 @@
...
+      if get_verbosity():
+        print_green('Found docker-compose.')
+      command_to_use = 'docker-compose'
+    except CalledProcessError:
+      if get_verbosity():
+        print_yellow(
+          'docker-compose command not found. Will use "docker compose" assuming it exists.'
+        )
+      command_to_use = 'docker compose'
+    return command_to_use
```
amonkhouse, line 25: mentioned elsewhere but I don't fully follow why this isn't included in `docker_compose_helpers.py`. I also think it could be named slightly better, something like `detect_compose_command`

hokusai/commands/development.py
```diff
@@ -1,107 +1,108 @@
...
-
-  shout("COMPOSE_COMPATIBILITY=true docker-compose -f %s -p hokusai stop" % docker_compose_yml, print_output=True)
+  compose_command = generate_compose_command(
+    filename,
+    default_yaml_file=DEVELOPMENT_YML_FILE
+  )
+  shout(
+    f'{compose_command} -p hokusai stop',
+    print_output=True
+  )
```
amonkhouse, line 52: these all look a lot better 👍

amonkhouse (changes requested):
in general this looks good to me but I think the circular import should be fully avoided as it makes it unclear what each file is for, I've left some comments about options I see

hokusai/services/docker.py
artsyjian: renamed function to `detect_compose_command`.

hokusai/lib/docker_compose_helpers.py
artsyjian: agreed. opted to move function to helpers.

amonkhouse (approved):
lgtm!

PR #439, opened 2025-09-10
feat: add confirmation prompt and --assume-yes flag to delete commands

dblandin:
Add interactive confirmation prompt to `hokusai production delete` and `hokusai staging delete` commands to prevent accidental deletion of Kubernetes resources. The prompt displays the project name, environment, and YAML file being used.

- Add --assume-yes/-y flag to bypass confirmation for automation
- Update integration test to use --assume-yes flag
- Confirmation message shows context: project, environment, and file path (if provided).

cc/ @brainbicycle 

**Example Invocation with Interactive Prompt**

Screenshot 2025-09-10 at 15 05 03

### DIFF
--- a/hokusai/lib/common.py
+++ b/hokusai/lib/common.py
@@ -2,6 +2,7 @@
 
 import os
 
+import click
 import json
 import platform
 import random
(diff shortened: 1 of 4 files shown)


hokusai/lib/common.py
```diff
@@ -380,3 +380,28 @@ def write_temp_file(data, dir):
 def yaml_content_with_header(content_string):
   ''' return content with yaml header prepended '''
   return YAML_HEADER + content_string
+
+def confirm_delete_resources(project_name, environment, yaml_file):
+  ''' Display confirmation prompt for destructive delete operations '''
+  import click
```
brainbicycle, line 386: is there a reason to inline the import here rather than put at the top of the file?

brainbicycle (approved):
small comment on the inlined import, thank you for adding!!!

hokusai/lib/common.py
dblandin: Ahh, good catch – I've gone ahead and moved it to the top of the file.

artsyjian (approved):
thanks!

PR #453, opened 2026-05-19
chore: upgrade boto3 and update tests for aws login credential support

dblandin:
## Context

`aws login` is [the modern alternative](https://docs.aws.amazon.com/signin/latest/userguide/command-line-sign-in.html) to managing static IAM access key credentials locally. It initiates a browser-based auth flow with MFA handled in the browser, then writes short-lived session credentials to `~/.aws/login/cache`. Sessions last ~12 hours. AWS has started actively nudging users toward this flow — it's surfaced in `aws configure` tips and in credential error messages.

When using `aws login`, hokusai currently fails with a credentials error. This PR upgrades boto3 to add support for the SSO credential provider so hokusai works with credentials obtained this way.

cc/ @ovasdi 

## Changes

- Upgrade boto3 from ^1.34.25 to ^1.43.10; narrow Python requirement to >=3.10 to match boto3's new minimum
- Add botocore[crt] (awscrt) dependency, required by the SSO credential provider
- Update ECR and docker-compose-helper unit tests: newer botocore calls the regional STS endpoint (sts.<region>.amazonaws.com) instead of the global one, causing httpretty mocks to miss; also mock get_region_name in test setUp so tests aren't sensitive to AWS_REGION being set in the environment

### Before

```shell
$ HOKUSAI_GLOBAL_CONFIG=s3://... hokusai configure --skip-kubectl
Error: Not able to load config from s3://...
botocore.exceptions.NoCredentialsError: Unable to locate credentials
```

After upgrading boto3, a secondary error surfaced requiring an additional dependency for the SSO credential provider:

```
botocore.exceptions.MissingDependencyException: Missing Dependency: Using the login
credential provider requires an additional dependency. You will need to pip install
"botocore[crt]" before proceeding.
```

### After

```shell
$ HOKUSAI_GLOBAL_CONFIG=s3://... hokusai configure --skip-kubectl
Downloading kubeconfig from s3://artsy-citadel/k8s/config-dev to ~/.kube ...
```


artsyjian:
it [appears](https://artsy.slack.com/archives/CA8SANW3W/p1779206895916939) that the discussion on migrating from long term keys (`~/.aws/credentials`) to `aws login` is still ongoing, so i think we shouldn't introduce any changes here that would disrupt our old workflow.

so if this change is completely backward compatible, meaning, long term keys will continue to work, then i am okay with merging.

artsyjian (approved):
thanks

PR #454, opened 2026-05-19
chore: manage dev toolchain and update setup docs with mise

dblandin:
## Summary

Replaces the pyenv + manual virtualenv + manual Poetry installation workflow with [mise](https://mise.jdx.dev/), which manages all three automatically.

### Before

Developers needed to:
1. Install pyenv with platform-specific build dependencies
2. Install the correct Python version via pyenv
3. Create a virtualenv with pyenv-virtualenv
4. Manually install Poetry via pip

### After

```
$ mise install
$ poetry install
```

## Changes

- Add .mise.toml pinning Python 3.10 and Poetry 1.2.2, with mise's virtualenv support configured to create and activate .venv automatically
- Update README to reflect the simplified setup, noting that the Makefile hokusai target and scripts/ci_macos_build_hokusai.sh are intentionally unchanged as they use pyenv in CI environments where mise is not available

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -1,0 +1,6 @@
+[tools]
+python = "3.10"
+poetry = "1.2.2"
+
+[env]
+_.python.venv = { path = ".venv", create = true }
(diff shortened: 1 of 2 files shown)


narikazu (approved):
Love it!

PR #455, opened 2026-05-19
feat: make kubectl-dir and kubectl-version optional in global config

dblandin:
## Summary

Previously, kubectl-dir and kubectl-version were required fields in ~/.hokusai.yml, even for users who manage kubectl externally (e.g. via mise or brew). This meant the config held stale metadata that hokusai never actually used.

## Changes
- kubectl-dir is now optional. When unset, hokusai resolves kubectl from PATH via shutil.which, raising a clear error if it can't be found.
- kubectl-version is now optional. It remains required only when kubectl-dir is set (i.e. when hokusai is managing the kubectl binary and needs to know what version to
  download).

### DIFF
--- a/hokusai/lib/global_config.py
+++ b/hokusai/lib/global_config.py
@@ -64,16 +64,18 @@ class HokusaiGlobalConfig:
   def validate_config(self):
     ''' sanity check config '''
     required_vars = [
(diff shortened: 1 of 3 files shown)


narikazu (approved):
LGTM

